PR #20, opened 2020-10-28
 Add a rake task that can be used for local development to setup an admin user

mejackreed:
Mostly cribbed from the approach in Spotlight: https://github.com/projectblacklight/spotlight/blob/master/lib/tasks/spotlight_tasks.rake

Added for local development convenience


cbeer (approved):

cbeer:
Good idea.

PR #23, opened 2020-10-28
Links user to organizations

hectorcorrea:
Adds link from Users to Organizations


hectorcorrea:
Not needed. This will be done through rolify.

Issue #4, opened 2020-10-27
Spike with thoughtbot/administrate (or something else?) for administrative actions (creating users / orgs / etc)


jkeck:
- Does administrate handle files well?
- Does administrate handle users + roles?

PR #28, opened 2020-10-28
[SPIKE] Add administrate

mejackreed:
Part of #4


mejackreed:
Closing this as I don't think we will end up going down this path.

PR #33, opened 2020-10-29
Add the ability to auth via a JWT

### LINKED ISSUE #7: Demonstrate file uploads using CURL

jkeck:
Closes #7

Added an example `curl` command in the upload pages that uses the generated JWT

Screen Shot 2020-10-29 at 2 43 30 PM


config/settings.yml
```diff
@@ -0,0 +1,3 @@
+jwt:
+  secret: 'i♥️dolphins'
```
jkeck, line 2: 🐬

app/models/ability.rb
```diff
@@ -4,8 +4,26 @@
...
+                          Settings.jwt.secret,
+                          Settings.jwt.algorithm
+                        )[0]
+                      else
+                        {}
+                      end
+
+      can :read, Organization, allowlisted_jwts: { jti: token_payload['jti'] }
+
+      can %i[create read update], [Stream, Upload, Batch], organization: { allowlisted_jwts: { jti: token_payload['jti'] } }
```
jkeck, line 23: Not sure what other permissions we might want in this area, but I imagine we can iterate.

mejackreed (approved):
Thanks I was able to get this going locally. Nice job on it

PR #39, opened 2020-10-30
Apply a little more style to the organization#show page

cbeer:
Screen Shot 2020-10-30 at 07 14 29


app/views/organizations/show.html.erb
```diff
@@ -1,18 +1,25 @@
-<h1><%= @organization.name %></h1>
+<div class="position-relative overflow-hidden mb-3 bg-light">
```
mejackreed, line 1: I'm good with this. Just pinning this idea for future, but maybe we want to make this a content_for as well as so other pages can easily tap into this design element also

PR #51, opened 2020-11-02
Use the updated_at timestamp of the token to determine if it has been used

jkeck:
Update that timestamp when a token has been used for access and display that information on the dashboard.

This updated_at strategy likely won't work if we add other data (like a name/label), but just thought I would throw this up to start the convo on how we might want to approach this.

Screen Shot 2020-11-02 at 10 00 04 AM

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -4,6 +4,7 @@
 class Ability
   include CanCan::Ability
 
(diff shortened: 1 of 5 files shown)


cbeer:
Seems reasonable to me 🤷‍♂️

PR #68, opened 2020-11-04
Handle race condition for unanalyzed files

mejackreed:


cbeer:
What's the race condition? That we're not able to read files until they go through the analyzers?

mejackreed:
Mentioned in #67 

> However, this may unveil a race condition not previously noticed. Whereas, the after_commit in Upload may enqueue itself and be ran before analysis has completed.

But maybe its not a problem especially if we don't depend on any analysis metadata. I thought I ran into an issue with this, but now I can't recreate it.

PR #69, opened 2020-11-04
Setup confirmable users

mejackreed:
Do we need to do additional things to get email going? Open to other alternatives here too

### DIFF
--- a/app/views/devise/shared/_links.html.erb
+++ b/app/views/devise/shared/_links.html.erb
@@ -1,0 +1,27 @@
+<div class="form-group">
+  <%- if controller_name != 'sessions' %>
+    <%= link_to t(".sign_in"), new_session_path(resource_name) %><br />
+  <% end -%>
+
+  <%- if devise_mapping.registerable? && controller_name != 'registrations' %>
+    <%= link_to t(".sign_up"), new_registration_path(resource_name) %><br />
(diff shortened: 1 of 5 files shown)


app/views/devise/shared/_links.html.erb
```diff
@@ -0,0 +1,27 @@
...
+  <%- if devise_mapping.registerable? && controller_name != 'registrations' %>
+    <%= link_to t(".sign_up"), new_registration_path(resource_name) %><br />
+  <% end -%>
+
+  <%- if devise_mapping.recoverable? && controller_name != 'passwords' && controller_name != 'registrations' %>
+    <%= link_to t(".forgot_your_password"), new_password_path(resource_name) %><br />
+  <% end -%>
+
+  <%- if devise_mapping.confirmable? && controller_name != 'confirmations' %>
+    <%= link_to t('.didn_t_receive_confirmation_instructions'), new_user_confirmation_path(resource_name) %><br />
```
mejackreed, line 15: An issue upstream here causing us to overwrite partial to use `new_user_confirmation_path` instead of `new_confirmation_path`

mejackreed: Or maybe its our issue to somehow fix in routes? 🤷‍♂️

PR #98, opened 2020-11-05
Produce full downloads for each organization weekly

### LINKED ISSUE #95: Schedule monthly full dumps of normalized records

cbeer:
Eventually, we might want to make this monthly, but for now run it frequently so we can see it in action more.

Fixes #95

### DIFF
--- a/config/schedule.rb
+++ b/config/schedule.rb
@@ -18,3 +18,12 @@
 # end
 
 # Learn more: http://github.com/javan/whenever
+
+# TODO: Raise this to monthly eventually?
+every :sunday do
+  runner 'GenerateFullDumpJob.enqueue_all'
+end
+
+every :day do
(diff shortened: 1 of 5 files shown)


config/schedule.rb
```diff
@@ -18,3 +18,8 @@
 # end
 
 # Learn more: http://github.com/javan/whenever
+
+# TODO: Raise this to monthly eventually?
+every :sunday do
+  runner 'DownloadJob.enqueue_all'
```
mejackreed, line 24: Could we comment this out for now? Or what until we expand our storage and account for that?

mejackreed (changes requested):
Quick request re: current storage constraints

cbeer:
I think this is ready for re-review now that we're compressing these dumps down to just a few GB per organization?

mejackreed (approved):

PR #102, opened 2020-11-05
Compress full-dumps using ruby Zlib.

### LINKED ISSUE #97: Investigate compressing full dumps
As long as we have no use-cases for reading back partial data from the normalized dumps, we might be able to compress the full + delta dumps (#87 + #96) when we create them.

https://ruby-doc.org/stdlib-2.7.0/libdoc/zlib/rdoc/Zlib/GzipWriter.html looks promising as long as we intend to produce a single file. If we need/want to chunk the records (for MARCXML in particular), can we do that easily?

jkeck:
Closes #97

### DIFF
--- a/app/jobs/download_job.rb
+++ b/app/jobs/download_job.rb
@@ -5,31 +5,30 @@
 class DownloadJob < ApplicationJob
   # rubocop:disable Metrics/AbcSize
   def perform(organization)
-    xmlfile = Tempfile.new("#{organization.slug}-marcxml")
-    xmlwriter = MARC::XMLWriter.new(xmlfile)
+    xmlfile = Tempfile.new("#{organization.slug}-marcxml", binmode: true)
     binary_file = Tempfile.new("#{organization.slug}-marc21", binmode: true)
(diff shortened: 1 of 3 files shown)


app/jobs/download_job.rb
```diff
@@ -6,14 +6,16 @@ class DownloadJob < ApplicationJob
...
     organization.default_stream.uploads.each do |upload|
       upload.each_marc_record_metadata.each do |record|
-        file.write(record.marc.to_marc)
+        compressor.write record.marc.to_marc
       end
     end
-    file.rewind
-    organization.full_dump_binary.attach(io: file, filename: "#{organization.slug}_#{Time.zone.today}")
+    compressor.close
+    organization.full_dump_binary.attach(io: File.open(file), filename: "#{organization.slug}_#{Time.zone.today}")
```
cbeer, line 16:
```suggestion
    organization.full_dump_binary.attach(io: File.open(file), filename: "#{organization.slug}_#{Time.zone.today}.gz")
```

cbeer (approved):

PR #133, opened 2020-11-09
Use case statement from metadata instead of inferring based off of ex…

mejackreed:
…tension

We may need to re-run analysis on prod data if we ship this

### DIFF
--- a/app/models/upload.rb
+++ b/app/models/upload.rb
@@ -21,11 +21,10 @@ class Upload < ApplicationRecord
     return to_enum(:each_marc_record_metadata) unless block_given?
 
     files.each do |file|
-      content_type = file.blob.content_type
-      filename = file.blob.filename.to_s
-      if content_type.ends_with?('xml') || filename.ends_with?('xml')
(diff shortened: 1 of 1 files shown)


app/models/upload.rb
```diff
@@ -21,11 +21,10 @@ def each_marc_record_metadata(&block)
     return to_enum(:each_marc_record_metadata) unless block_given?
 
     files.each do |file|
-      content_type = file.blob.content_type
-      filename = file.blob.filename.to_s
-      if content_type.ends_with?('xml') || filename.ends_with?('xml')
+      case file.blob.metadata['analyzer']
```
cbeer, line 24: It's possible the analyzer hasn't finished yet when this process kicks off, I think? Should we blow up and wait for the analyzer or share some common file identification logic?

mejackreed: Yep i guess that definitely is possible. 🤷 blowing up can work. I could combine some of this? https://github.com/ivplus/aggregator/pull/68/files

Thoughts?

PR #158, opened 2020-11-11
Set marc21 and marcxml mimetypes for these files Fixes #128

### LINKED ISSUE #128: Update upload content types to use MARC-specific mime types if/when we can detect that
Uploaded MARC files come in with generic mime-types. If we can figure out what they are, these mime types would be better:

MARC21: `application/marc`
MARCXML: `application/marcxml+xml`

mejackreed:
I left `marcxml_gzip` and `marc21_gzip` alone for now as it seems useful to know that?

### DIFF
--- a/app/analyzers/marc_analyzer.rb
+++ b/app/analyzers/marc_analyzer.rb
@@ -16,9 +16,7 @@ class MarcAnalyzer < ActiveStorage::Analyzer
     { analyzer: self.class.to_s, valid: false, error: e.message }
   end
 
-  private
-
   def reader
-    @reader ||= MarcRecordService.new(@blob)
+    @reader ||= MarcRecordService.new(blob)
   end
 end
--- a/lib/active_storage_blob_content_types.rb
+++ b/lib/active_storage_blob_content_types.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+##
+# An extension that will be included into the ActiveStorage::Blob class
+module ActiveStorageBlobContentTypes
+  def analyze
+    update! metadata: metadata.merge(extract_metadata_via_analyzer), content_type: marc_content_type
+  end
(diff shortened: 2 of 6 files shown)


app/analyzers/marc_analyzer.rb
```diff
@@ -7,6 +7,11 @@ def self.accept?(blob)
     MarcRecordService.new(blob).identify != :unknown
   end
 
+  def initialize(blob)
+    super(blob)
+    blob.update_content_type(reader.identify)
```
cbeer, line 12: I'm a little wary of doing database updates in the initializer.. but we also seem kinda constrained by the API on offer. I'm going to make a suggestion in `ActiveStorageBlobContentTypes`, but I'm not sure it's any better really..

lib/active_storage_blob_content_types.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+##
+# An extension that will be included into the ActiveStorage::Blob class
+module ActiveStorageBlobContentTypes
```
cbeer, line 5: What if we.... override `analyze` to make the content type update?

https://github.com/rails/rails/blob/master/activestorage/app/models/active_storage/blob/analyzable.rb

So, something like...

```
def analyze
  update! metadata: metadata.merge(extract_metadata_via_analyzer), content_type: marc_content_type
end

def marc_content_type
  case analyzer.identify
  when :marc21 then 'application/marc'
  when :marcxml then 'application/marcxml+xml'
  else content_type
  end
end
```

?

app/analyzers/marc_analyzer.rb
mejackreed: Thanks for the review, and yes I agree with for sure. I've pushed an update.

PR #160, opened 2020-11-12
Allow file upload via URL

### LINKED ISSUE #159: "Upload" files by URL

jkeck:
Closes #159 

Screen Shot 2020-11-11 at 5 19 38 PM

### DIFF
--- a/app/controllers/uploads_controller.rb
+++ b/app/controllers/uploads_controller.rb
@@ -76,6 +76,6 @@ class UploadsController < ApplicationController
 
   # Only allow a list of trusted parameters through.
   def upload_params
-    params.require(:upload).permit(:name, files: []).tap { |p| p['files'].reject!(&:blank?) }
+    params.require(:upload).permit(:name, :url, files: []).tap { |p| p['files']&.reject!(&:blank?) }
   end
 end
--- a/app/jobs/attach_remote_file_to_upload_job.rb
+++ b/app/jobs/attach_remote_file_to_upload_job.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+##
+# Job to upload files from a URL in the background
+class AttachRemoteFileToUploadJob < ApplicationJob
+  queue_as :default
+
+  def perform(upload)
+    upload.files.attach(io: URI.open(upload.url), filename: upload.name)
+  rescue SocketError, OpenURI::HTTPError => e
+    error = "Error opening #{url}: #{e}"
+    Rails.logger.info(error)
+    Honeybadger.notify(error)
+    raise e
+  end
+end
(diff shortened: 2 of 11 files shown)


app/controllers/uploads_controller.rb
```diff
@@ -76,6 +76,6 @@ def create_params
 
   # Only allow a list of trusted parameters through.
   def upload_params
-    params.require(:upload).permit(:name, files: []).tap { |p| p['files'].reject!(&:blank?) }
+    params.require(:upload).permit(:name, :url, files: []).tap { |p| p['files']&.reject!(&:blank?) }
```
jkeck, line 79: We could technically take an array of URLs but 🤷

app/jobs/attach_remote_file_to_upload_job.rb
```diff
@@ -0,0 +1,16 @@
...
+class AttachRemoteFileToUploadJob < ApplicationJob
+  queue_as :default
+
+  def perform(upload)
+    upload.files.attach(io: URI.open(upload.url), filename: upload.name)
+  rescue SocketError, OpenURI::HTTPError => e
+    error = "Error opening #{url}: #{e}"
+    Rails.logger.info(error)
+    Honeybadger.notify(error)
+    raise e
```
jkeck, line 14: I'm just re-raising here so we retry (incase the issue was a temp. network blip or the external file store is temp. responding w/ an error).

Not sure if we want to do anything more sophisticated now, or wait until we have more data on what kind of issues we're likely to see uploading via URL.

app/models/upload.rb
```diff
@@ -6,6 +6,11 @@ class Upload < ApplicationRecord
   belongs_to :stream, touch: true
   has_one :organization, through: :stream
   has_many :marc_records, dependent: :delete_all
+  validates :url, presence: true, if: proc { |upload| upload.files.blank? }
+  validates :files, presence: true, if: proc { |upload| upload.url.blank? }
```
jkeck, line 10: It feels like there should be a better way to do this.

app/models/upload.rb
```diff
@@ -6,6 +6,11 @@ class Upload < ApplicationRecord
...
+  validate :valid_url, if: proc { |upload| upload.url.present? }
+
+  after_create :attach_file_from_url
```
jkeck, line 13: I'm doing this in an `after_create`, which begs the question what editing/updating an upload should do.  I can change this to `after_save` and allow URLs to be re-uploaded, but not sure if that's the desired behavior or not (and if so, would we want to check if the URL had changed before re-downloading so that a change to the name doesn't trigger re-downloading the file).

cbeer:
I'm good with this for now; if we discover additional needs later we can address it then.

PR #180, opened 2020-11-12
Adds User and ipaddress to Upload fixes #179

### LINKED ISSUE #179: Record some additional metadata with uploads for ease of debugging (which user? maybe IP information?)

mejackreed:


app/models/upload.rb
```diff
@@ -6,6 +6,7 @@ class Upload < ApplicationRecord
   belongs_to :stream, touch: true
   has_one :organization, through: :stream
   has_many :marc_records, dependent: :delete_all
+  belongs_to :user, optional: true
```
mejackreed, line 9: Should this belong to any `User` who touches it?

app/models/user.rb
```diff
@@ -10,6 +10,8 @@ class User < ApplicationRecord
          :recoverable, :rememberable, :validatable,
          :invitable, :confirmable
 
+  has_many :uploads, dependent: :destroy
```
mejackreed, line 13: Should we do this?

cbeer, line 13:
```suggestion
  has_many :uploads
```

I don't think we want to remove the upload if the user is deleted (for whatever reason... they leave the org or something) ((although I think rubocop will have something to say about this...))

app/controllers/uploads_controller.rb
```diff
@@ -71,7 +71,11 @@ def current_stream
   end
 
   def create_params
-    upload_params.merge(stream_id: current_stream.id)
+    upload_params.merge(
+      stream_id: current_stream.id,
+      user_id: current_user&.id,
+      ip_address: request.env['HTTP_X_FORWARDED_FOR'] || request.remote_ip
```
cbeer, line 77: Is it also possible to track which token was used, if any?

mejackreed: Sure thing!

app/models/upload.rb
cbeer: If the files change, I guess? The goal with #179 is to give us some clues if/when users report upload trouble or something.

app/models/user.rb
mejackreed: I've added this in a commit.

PR #181, opened 2020-11-12
Try to get the filename from the ContentDisposition header from…

### LINKED ISSUE #174: Use the `Content-Disposition` filename attribute as the file name when uploading by URL
Eg. from box:

```
13:10 % curl -I '....'
HTTP/1.1 200 OK
...
Content-Disposition: attachment;filename="chicago-full-2020-11-07.xml.gz";filename*=UTF-8''chicago-full-2020-11-07.xml.gz
```

We'd use the file name `chicago-full-2020-11-07.xml.gz`.

jkeck:
…URL based uploads.

Closes #174

### DIFF
--- a/app/jobs/attach_remote_file_to_upload_job.rb
+++ b/app/jobs/attach_remote_file_to_upload_job.rb
@@ -6,11 +6,26 @@ class AttachRemoteFileToUploadJob < ApplicationJob
   queue_as :default
 
   def perform(upload)
-    upload.files.attach(io: URI.open(upload.url), filename: upload.name)
+    io = URI.open(upload.url)
+    upload.files.attach(io: io, filename: filename_from_io(io) || upload.name)
   rescue SocketError, OpenURI::HTTPError => e
     error = "Error opening #{url}: #{e}"
     Rails.logger.info(error)
     Honeybadger.notify(error)
     raise e
+  end
+
+  private
+
+  def filename_from_io(io)
+    content_disposition = io.meta['content-disposition']&.split(';') || []
+    filenames = content_disposition.map do |disp|
+      key, value = disp.split('=')
+      next unless key == 'filename'
+
+      value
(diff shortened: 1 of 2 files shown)


spec/jobs/attach_remote_file_to_upload_job_spec.rb
```diff
@@ -1,12 +1,29 @@
 # frozen_string_literal: true
 
 require 'rails_helper'
+require 'delegate'
+
+class RemoteUploadFixture < SimpleDelegator
```
jkeck, line 6: LOL @ RSPEC

lol

app/jobs/attach_remote_file_to_upload_job.rb
```diff
@@ -6,11 +6,26 @@ class AttachRemoteFileToUploadJob < ApplicationJob
...
   rescue SocketError, OpenURI::HTTPError => e
     error = "Error opening #{url}: #{e}"
     Rails.logger.info(error)
     Honeybadger.notify(error)
     raise e
   end
+
+  private
+
+  def filename_from_io(io)
```
cbeer, line 20: Who woulda thought that parsing HTTP headers in Rails would be... this... interesting..

jkeck: To be fair, I haven't actually tried this with something that has a content-disposition but nothing I've ready leads me to believe rails does anything great for us here (although apparently mechanize does have some niceties around this, but that method is, uh, [something](https://github.com/sparklemotion/mechanize/blob/b959fb10d2e5943e69c268993481a1b73c80717e/lib/mechanize/parser.rb#L106) [else](https://github.com/sparklemotion/mechanize/blob/b959fb10d2e5943e69c268993481a1b73c80717e/lib/mechanize/http/content_disposition_parser.rb))

Issue #176, opened 2020-11-12
[Spike] expose direct access to MARC/MARCXML records by organization + 001 field


cbeer:
Related to #172 (and maybe simpler..). We can find out where to find a record is given the organization and 001 value, so let's expose an API to get those records back for debugging, development (fixtures?) and other adminstration tasks

mejackreed:
Closing this as implemented in #185 but perhaps there are more things to consider.

PR #185, opened 2020-11-13
Starts to expose an API for accessing MarcRecords

mejackreed:
Part of #176

 - [x] Offer HTML views
 - [x] Link from record counts

Screen Shot 2020-11-16 at 4 03 11 PM

### DIFF
--- a/app/controllers/marc_records_controller.rb
+++ b/app/controllers/marc_records_controller.rb
@@ -8,5 +8,12 @@ class MarcRecordsController < ApplicationController
 
   def show; end
 
-  def index; end
+  def index
+    @marc_records = @marc_records.where(marc001: index_params[:marc001]) if index_params[:marc001]
+    @marc_records = @marc_records.page(index_params[:page])
+  end
+
+  def index_params
+    params.permit(:page, :marc001)
+  end
 end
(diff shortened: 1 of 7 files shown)


app/controllers/marc_records_controller.rb
```diff
@@ -0,0 +1,12 @@
...
+
+# Controller to handle MarcRecords
+class MarcRecordsController < ApplicationController
+  load_and_authorize_resource :organization
+  load_and_authorize_resource through: :organization
+  protect_from_forgery with: :null_session, if: :jwt_token
+
+  def show; end
+
+  def index; end
```
mejackreed, line 11: Do we want to enable querying by a given `marc001`? By the latest of that `marc001` (or all unique records)?

mejackreed: Or enable pagination?

cbeer: Yes, I think so... and there's maybe use cases for showing either the latest or the record from a particular file (certainly the latter for debugging data errors).

app/views/marc_records/index.json.jbuilder
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+json.array! @marc_records do |marc_record|
```
cbeer, line 3: This is 50+ million records in production right now; we probably need to start with a paginated response instead of adding it later.

app/controllers/marc_records_controller.rb
mejackreed: I've added support for query by `marc001` and pagination. Lets circle about latest/sorting discussion as I think I want to better understand that.

PR #194, opened 2020-11-13
Add a form for organizations to provide item-level mapping information

cbeer:
Part of #162

Screen Shot 2020-11-13 at 15 26 53

Augmented MARC records append the normalized fields (in this case, a set of 999s with the mapped subfields), plus some provenance information using the `$8` tag (🤷‍♂️) to connect the POD normalized form back to the original data field.

### DIFF
--- a/app/services/augment_marc_record_service.rb
+++ b/app/services/augment_marc_record_service.rb
@@ -1,0 +1,52 @@
+# frozen_string_literal: true
+
+# Service class for augmenting source MARC records with POD information
+class AugmentMarcRecordService
+  attr_reader :steps, :organization
+
+  DEFAULT_STEPS = %i[
+    append_pod_provenance_information
+    append_normalized_item_information
+  ].freeze
+
+  def initialize(steps: nil, organization: nil)
+    @steps = steps || DEFAULT_STEPS
+    @organization = organization || Organization.new
+  end
+
+  def execute(source_record)
+    steps.each_with_object(duped_record(source_record)) do |step, record|
+      send(step, record)
+    end
+  end
+
+  private
+
+  def append_pod_provenance_information(record)
+    return if organization.code.blank?
+
+    record.append(MARC::DataField.new('900', nil, nil, ['b', organization.code]))
+  end
+
+  def append_normalized_item_information(record)
+    return if organization.normalization_steps.blank?
+
+    organization.normalization_steps.each_value do |step|
+      map_normalized_item_field(record, step)
+    end
+  end
+
+  def map_normalized_item_field(record, step)
+    record.fields(step['source_tag']).each do |field|
+      subfield_data = step['subfields'].sort_by { |k, _v| k }.map do |destination_subfield, source_subfield|
+        [destination_subfield, field[source_subfield]] if source_subfield && field[source_subfield].present?
+      end.compact
+
+      record.append(MARC::DataField.new(step['destination_tag'], nil, nil, *subfield_data))
+    end
+  end
+
+  def duped_record(source_record)
+    MARC::Record.new_from_hash(source_record.to_hash)
+  end
+end
--- a/app/services/augment_marc_record_service.rb
+++ b/app/services/augment_marc_record_service.rb
@@ -14,6 +14,7 @@ class AugmentMarcRecordService
     @organization = organization || Organization.new
   end
 
+  # @param [MARC::Record] record to modify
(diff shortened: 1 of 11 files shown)


app/services/augment_marc_record_service.rb
```diff
@@ -0,0 +1,68 @@
...
+
+    organization.normalization_steps.each_value do |step|
+      map_normalized_item_field(record, step)
+    end
+  end
+
+  def map_normalized_item_field(record, step)
+    record.fields(step['source_tag']).each.with_index do |field, index|
+      subfield_data = map_subfield_data(field, step['subfields'])
+      field_link = ['8', "#{field.tag}.#{index}\\p"]
```
jkeck, line 49: I'm not sure if it matters much, but the typical formatting of $8 would be something like `999-01`, `999-02`, etc. 

I suppose we can't count on there only being 100 fields (for the padded zeros) but we might want to consider using a hyphen to concatenate the tag and the index to be consistent w/ other MARC fields.

jkeck: I'm not familiar w/ the `/p` what does that signify?

app/views/organizations/_form.html.erb
```diff
@@ -14,6 +14,17 @@
     </label>
   </div>
 
+  <% if organization.persisted? %>
+    <hr />
+
+    <h3>Item information</h3>
+    <%= form.hidden_field 'normalization_steps[0][destination_tag]', maxlength: 3, value: organization.normalization_steps.values.dig(0, 'destination_tag') || 998 %>
```
jkeck, line 21: Any particular reason we're making this a hidden field?  Maybe it could be a disabled text input?  🤷

app/views/organizations/_form.html.erb
```diff
@@ -14,6 +14,17 @@
...
+    <%= form.text_field 'normalization_steps[0][source_tag]', maxlength: 3, value: organization.normalization_steps.values.dig(0, 'source_tag'), label: 'What MARC field contains item-level information?' %>
+    <%= form.text_field 'normalization_steps[0][subfields][i]', maxlength: 1, value: organization.normalization_steps.values.dig(0, 'subfields', 'i'), label: 'Local identifier subfield' %>
+    <%= form.text_field 'normalization_steps[0][subfields][a]', maxlength: 1, value: organization.normalization_steps.values.dig(0, 'subfields', 'a'), label: 'Call number subfield' %>
+    <%= form.text_field 'normalization_steps[0][subfields][m]', maxlength: 1, value: organization.normalization_steps.values.dig(0, 'subfields', 'm'), label: 'Local library' %>
```
jkeck, line 25: I recall pain around serialized data, form names, and URL params in the past, is that why we're jumping through the extra index in the json here (or are we doing this to leave ourselves open to adding more steps)?

jkeck (approved):
A few comments here, nothing that I think is blocking at all.  Feel free to ignore if you like.

app/services/augment_marc_record_service.rb
cbeer: Hm. I made that linking field up based on the the MARC21 spec that says:
> $8 [linking number].[sequence number]\[field link type]

and 
> p - Metadata provenance  Used in a record to link a field with another field containing information about provenance of the metadata recorded in the linked field.
(which.. is maybe technically wrong, but seemed more appropriate than other choices)

app/views/organizations/_form.html.erb
cbeer: @mejackreed and I went back and forth on what to do with these fields and whether to add some abstraction in the form (that we then map back to MARC fields in the app) or just expose the raw MARC mappings and call it good enough for now.

So, 🤷‍♂️ ? I'm happy to go either way.

app/views/organizations/_form.html.erb
cbeer: Yep, both. This way seemed to play nice with the json column type and rails, so we backed away slowly.

app/services/augment_marc_record_service.rb
jkeck: Ah, I'm thinking about $6 actually, which uses `-` instead of `.` because of course it does.

app/views/organizations/_form.html.erb
jkeck: The 🤷s have it

PR #202, opened 2020-11-17
[Spike] displaying some marc profiling data

mejackreed:
This is just a proof of concept. What is a little unclear to me is what is useful to see here?

Screen Shot 2020-11-17 at 5 15 20 AM

### DIFF
--- a/app/controllers/charts_controller.rb
+++ b/app/controllers/charts_controller.rb
@@ -11,4 +11,11 @@ class ChartsController < ApplicationController
   def records
     render json: Statistic.where(resource_type: 'Organization').group_by_day(:created_at).sum(:unique_record_count)
   end
+
+  def orgs
+    authorize!(:show, :organization)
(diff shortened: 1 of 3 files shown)


cbeer:
I'll offer some things we might learn from the profiles below (and i'm happy for them to get pushed into separate tickets or 🤷‍♂️ ; I'm also guessing at some MARC fields right now, more than happy for someone to edit them if they get to it before I do):

- basic record validation (we're assuming every record has a `001` field; is this actually true? (and I suspect there are some other specific fields or subfields we could ask about... oclc number, ISBN, etc?)
- do the records contain MHLD information (85x? 86x?) 
- do the records contain multilingual data (88x?)
- what non-standard fields + subfields are people using? (https://github.com/pkiraly/metadata-qa-marc/blob/master/src/main/resources/marc-schema.json might be useful?)

cbeer:
Superceded by #211, i think?

Issue #175, opened 2020-11-12
Add some data retention policies so storage doesn't grow out of hand


cbeer:
Given that:
- we don't need/want to keep all the uploads ever
- we do want to keep track of the marc records that have previously appeared (so we can provide deletes for them.. and do we need to provide these indefinitely, or, say a year's worth.. and if we're throwing away the original files, where/how do we track these?)

The POD tech call was ok setting some limits on what we hold on to. Here's a first proposal:

- [x] Keep the most recent ("default") stream from each organization
- [x] Keep older streams for an organization for a quarter after it is no longer the default.
- [x] Keep the 2 most recent normalized dumps
- [ ] Keep deltas for an additional month (1 whole quarter)
- [ ] Keep deleted record information for 6 months after it is removed

PR #203, opened 2020-11-17
Add partial implementation of data retention policies part of #175

mejackreed:
I think I may need some more understanding about the last two:

```
# Keep deltas for an additional month (1 whole quarter)
# Keep deleted record information for 6 months after it is removed
```

This also adds a default value for the default column on streams so that we don't have the `true`, `nil`, `false` situation.

### DIFF
--- a/app/jobs/cleanup_and_remove_data_job.rb
+++ b/app/jobs/cleanup_and_remove_data_job.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+# Cleanup data per data retention policy
+class CleanupAndRemoveDataJob < ApplicationJob
+  def self.enqueue_all
+    Organization.find_each { |org| CleanupAndRemoveDataJob.perform_later(org) }
+  end
+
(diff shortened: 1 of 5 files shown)


cbeer:
We don't have deltas yet, so there's nothing to be done yet.

For "deleted record information", I think things like our `MarcRecord` instances qualify (I'm not sure if Rails is going to remove them automatically on us (and we'll have to do something to stop it) or if we need to sweep them up ourselves too)

mejackreed:
@cbeer I've confirmed the fact that our `MarcRecord` are gone on stream deletion. Have you put much thought in to how the reaping might work? It seems like we might have some UI consistancy issues. 

Two alternatives I've thought of:
 - I wonder if a stream could be "archived" which might put it into a state where data cleanup is more explicitly understood
 - Or the easy thing might just be change streams to 6 months to sync w/ our MarcRecord needs and call it a day.

cbeer:
options 1 or 2 sound fine to me. Option 1 might be easy enough (flag it, purge the files)?

PR #214, opened 2020-11-18
Add some plumbing to Hydrate the Data Lake UI (for dev purposes)

### LINKED ISSUE #177: Provide an easy way to seed data for local development using fixtures (or harvest some content from pod.stanford.edu?)
This might have some relation to #172 / #176, if those endpoints are performant enough to get some data (and we had some way to identify records)

(Note: Harvesting content direct from POD may have some additional utility proving that data can be synced from POD for other purposes too, so it should consume our "public" APIs. )

jkeck:
Closes #177 

This PR also addresses allowing download (and all) scoped tokens to read/download any organization's data ([card](https://github.com/ivplus/aggregator/projects/1#card-49541698))

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -5,3 +5,27 @@
 #
 #   movies = Movie.create([{ name: 'Star Wars' }, { name: 'Lord of the Rings' }])
 #   Character.create(name: 'Luke', movie: movies.first)
+
+require_relative '../spec/support/marc_fixture_seed_fetcher'
+
+Upload.skip_callback(:save_commit, :after, :perform_extract_marc_record_metadata_job)
+
+Settings.marc_fixture_seeds.organizations.each do |org|
+  organization = Organization.create(name: org, slug: org.downcase)
+
+  MarcFixtureSeedFetcher.fetch_uploads(org.downcase) do |upload, files|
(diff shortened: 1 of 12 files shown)


db/seeds.rb
```diff
@@ -5,3 +5,27 @@
...
+
+Upload.skip_callback(:commit, :after, :perform_extract_marc_record_metadata_job)
+
+Settings.marc_fixture_seeds.organizations.each do |org|
+  organization = Organization.create(name: org, slug: org.downcase)
+
+  MarcFixtureSeedFetcher.fetch_uploads(org.downcase) do |upload, files|
+    upload = organization.default_stream.uploads.build(name: upload['name'])
+    files.map do |file|
+      io = URI.open(file['download_url'], 'Authorization' => "Bearer #{Settings.marc_fixture_seeds.token}")
```
jkeck, line 19: I could theoretically merge this into the file data in `MarcFixtureSeedFetcher` so we don't have to have knowledge of how to do authorization here, but 🤷

jkeck:
If this ends up looking good, might end up massaging the seed code a bit after we deploy this to prod and I use this with real production data (I could also deploy this branch and start trying it out before shipping if that seems warranted).

PR #218, opened 2020-11-18
Refactor MARC21 combining logic up to MarcRecordService so profiling …

cbeer:
…can use it too


jkeck (approved):
This looks good to me (for a PR review past 4pm), I'll leave the review open though in case anybody (w/ fresher eyes) wants to take a look too.

PR #229, opened 2020-11-19
Add a feature spec for downloading generated full dump files

jkeck:


app/controllers/proxy_controller.rb
```diff
@@ -17,7 +17,7 @@ class ProxyController < ActiveStorage::BaseController
   # rubocop:disable Metrics/AbcSize
   def show
     attachment = ActiveStorage::Attachment.find(params[:id])
-    authorize!(:read, attachment.record)
+    authorize!(:read, attachment.record.stream)
```
jkeck, line 20: I think this should be legit given that all our records are associated w/ streams...

spec/features/normalized_download_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+    user.add_role :member, organization
+    login_as(user, scope: :user)
+
+    FactoryBot.create_list(:upload, 2, :binary_marc, organization: organization, stream: organization.default_stream)
+    FactoryBot.create_list(:upload, 2, :binary_marc_gz, organization: organization, stream: organization.default_stream)
+  end
+
+  describe 'Full dumps' do
+    before do
+      allow(Time.zone).to receive(:today).and_return('2020-01-01')
```
jkeck, line 19: simpler times

PR #230, opened 2020-11-19
Add a public flag to organizations and use that to authorized access

cbeer:
And tweak #229 to use the flag to authorize access to attachments explicitly

This should also let us hide any of our test orgs from normal users

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -31,27 +31,37 @@ class Ability
       case @allowlisted_jwt.scope
       when 'all'
         can %i[create update], [Stream, Upload], organization: { allowlisted_jwts: { jti: token_payload['jti'] } }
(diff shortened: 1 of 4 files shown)


app/models/ability.rb
```diff
@@ -31,27 +31,37 @@ def initialize(user, token = nil)
...
+        can :read, [Stream, Upload], organization: { public: true }
+
+        can :read, ActiveStorage::Attachment, { record: { organization: { public: true } } }
       end
 
       allowlisted_jwt&.update(updated_at: Time.zone.now)
       return
     end
 
+    can :read, ActiveStorage::Attachment, { record: { organization: { public: true } } }
```
cbeer, line 51: I don't know if it's funny to write authz rules based on Rails-provided models, but 🤷‍♂️

PR #231, opened 2020-11-19
Allow POD provenance info to be provided for records where the …

jkeck:
…Organization hasn't provided their code.

### DIFF
--- a/app/services/augment_marc_record_service.rb
+++ b/app/services/augment_marc_record_service.rb
@@ -27,9 +27,9 @@ class AugmentMarcRecordService
   # provided the record to POD.
   # @param [MARC::Record] record to modify
   def append_pod_provenance_information(record)
-    return if organization.code.blank?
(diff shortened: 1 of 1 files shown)


app/services/augment_marc_record_service.rb
```diff
@@ -27,9 +27,9 @@ def execute(source_record)
   # provided the record to POD.
   # @param [MARC::Record] record to modify
   def append_pod_provenance_information(record)
-    return if organization.code.blank?
```
jkeck, line 30: Not sure if we intended to gate POD's provenance info from being added based on the Org adding a code or not (which is the behavior we we're currently experiencing)

cbeer: 👍 WFM. The initial thought was if we don't have a code, there's no reason to inject data... but knowing the record came from POD at all is probably useful regardless

PR #233, opened 2020-11-19
Resolves all remaining data retention issues

### LINKED ISSUE #175: Add some data retention policies so storage doesn't grow out of hand

mejackreed:
Fixes #175


cbeer:
Bold claim.

mejackreed:
Go **Bold** or Go Home

Issue #196, opened 2020-11-16
Collect download statistics / analytics


anarchivist:
- (define what statistics to collect)

cbeer:
We're currently collecting this information using ahoy:

```
001 > e = Ahoy::Event.last
#<Ahoy::Event id: 11, visit_id: 8363, user_id: 11, name: "Download", properties: {"attachment_id"=>3175, "attachment_name"=>"files", "byte_size"=>3073505837, "content_type"=>"application/zip", "filename"=>"penn_02DEC2020.zip", "organization_id"=>"penn"}, time: "2020-12-02 14:46:11"> 

002 > e.visit
#<Ahoy::Visit id: 8363, visit_token: "753b6b43-3ba1-48d2-b5f8-e35ef568ca5d", visitor_token: "391dd96f-d339-4721-b0f2-b2365371240f", user_id: 11, ip: "100.11.141.0", user_agent: "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) Ap...", referrer: "https://pod.stanford.edu/site_users", referring_domain: "pod.stanford.edu", landing_page: "https://pod.stanford.edu/organizations", browser: "Chrome", os: "Mac", device_type: "Desktop", country: nil, region: nil, city: nil, latitude: nil, longitude: nil, utm_source: nil, utm_medium: nil, utm_term: nil, utm_content: nil, utm_campaign: nil, app_version: nil, os_version: nil, platform: nil, started_at: "2020-12-02 14:45:17", token_id: nil, organization_id: "penn"> 
```

The Ahoy::Event has information about what file was downloaded, the org the file belonged to, and some stats about the file. The associated visit has the user, token, and relevant organization of who downloaded the data.

PR #237, opened 2020-11-20
Track download analytics

cbeer:
🤷‍♂️ 

Related to #196

### DIFF
--- a/spec/features/normalized_download_spec.rb
+++ b/spec/features/normalized_download_spec.rb
@@ -56,5 +56,25 @@ RSpec.describe 'Downloading normalzed files from POD', type: :feature do
         expect(marc['900']['b']).to eq organization.code
       end
     end
+
+    it 'tracks the download' do
+      visit organization_url(organization)
+
+      expect do
+        # Note, this won't work in a driver other that Rack::Test w/o some other magic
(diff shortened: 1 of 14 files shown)


spec/features/normalized_download_spec.rb
```diff
@@ -56,5 +56,25 @@
...
+    it 'tracks the download' do
+      visit organization_url(organization)
+
+      expect do
+        # Note, this won't work in a driver other that Rack::Test w/o some other magic
+        click_link "#{organization.slug}-2020-01-01-marc21.mrc.gz"
+      end.to change(Ahoy::Event, :count).by(1)
+
+      expect(Ahoy::Event.last.properties.with_indifferent_access).to include(
+        attachment_name: 'full_dump_binary',
```
mejackreed, line 69: Is there a way to query this somehow through the expected attachment/file ?

Or can we only do this from the visit itself?

cbeer: I don't think ahoy gives us anything for free, but I /think/ we can use the postgres json query stuff to dig it out easily (and, if not... just run a daily statistics report and call it good? 🤷‍♂️).

mejackreed: WFM

PR #239, opened 2020-11-20
Fix seed task skip_callback (and add some logging)

jkeck:


db/seeds.rb
```diff
@@ -8,24 +8,30 @@
 
 require_relative '../spec/support/marc_fixture_seed_fetcher'
 
-Upload.skip_callback(:save_commit, :after, :perform_extract_marc_record_metadata_job)
+Upload.skip_callback(:commit, :after, :perform_extract_marc_record_metadata_job)
```
jkeck, line 11: I guess I should have tried this before actually making the change 😅

cbeer: Where's the fun in that?

PR #242, opened 2020-11-21
Allow MarcRecords to be looked up via ISBN

jkeck:
I'm just pushing this up because I'll be off for a few days.  I don't think this is really tractable in its current path, but maybe somebody could take it and run.

Also don't know if this would be totally useless with #241  🤷 

One major hurdle this would have is that records have multiple ISBNs, so we would probably need a separate table for those so we could query effectively on which records had that particular ISBN (which maybe make the graphql option even more attractive).


mejackreed:
This is really cool! Does it make sense to also add this query ability to the `MarcRecordsController` and / or display the ISBN in the existing views?

PR #281, opened 2020-12-09
Gather code coverage data when running specs.

jkeck:


cbeer:
Actually using the gem? what a concept!

PR #467, opened 2022-03-11
Reorganize user options into nav dropdown menu

### LINKED ISSUE #446: Update user login logic in topbar
Currently POD users log in to the application, get to their user-related tasks (name, title, password, cancel account), and log out of the application using distinct links in the application topbar:

Screen Shot 2022-03-04 at 10 51 18 AM

## Current problem
This approach poses several problems:
- When you've provided a name as part of your user registration and then log in, the display showing you're logged in can be quite long, and since we also display a Logout link for logged in users, these links take up enough significant space in the topbar, as indicated by the red arrow in the screenshot above (already a problem on smaller viewports). This approach would become even more of a problem if in the future we want to add more options to the topbar (e.g., a Feedback button).
- Aside from potential space constraints, it isn't a very attractive way to present user-related options. 
- It's not especially obvious that selecting your name/email address is how you get to your user edit page with the options to edit your profile info, change your password, or delete your account.

## Proposed improvement
- No change to non-logged in approach; we can continue to show a Login link to users who are not yet logged in to POD.

- When user is logged-in, we replace the Login link with the user's email address and a select menu displaying all appropriate user-related options (based on the privileges of the user), including the Logout link.

This approach also provides flexibility in the future, if and when additional user-oriented options are added to POD, because we have a place to put the new affordances without having to find space in the topbar or other location. 

If I understand things correctly, we currently have four possible types of registered users:
- Registered on their own, so no organization association
- Registered via organization invite, so associated with that organization
- Registered via organization invite, and made a POD admin, so has both an organization association and admin privileges
- Registered on their own, so no organization association, but made a POD admin so has admin privileges (this might be an unintended case, but is currently possible)

The mockups below show how the user select menu would look in each of those cases.

### Example: No organization association
- The Edit profile option goes to `https://pod.stanford.edu/users/edit` 
- The Logout option serves the same purpose as the current Logout link in the topbar

Screen Shot 2022-03-04 at 10 53 04 AM

---

### Example: Organization association
- When the user is associated with an organization, we add an additional option to the menu. In the example below the user is associated with Duke, so the option is labeled `Duke POD` (we might be able to come up with a better label). This option would take the user to Duke's organization page (`https://pod.stanford.edu/organizations/duke`), just as the link under the "My organizations" heading on the homepage currently does.

Screen Shot 2022-03-04 at 10 54 03 AM

---

### Example: Organization association and POD admin
- When the user is a POD admin, we display the additional option `Manage users` in the menu. This goes to `https://pod.stanford.edu/site_users`, just as the "Users" link admin's see in the topbar currently does. 
- Importantly, adding this option to the user menu enables us to remove the "Users" link from the topbar, where it currently takes up space even though it is likely very rarely used.

Screen Shot 2022-03-04 at 12 22 33 PM

---

### Example: No organization association but POD admin
- Again, I'm not sure this is a case we want to have, but if a user is a POD admin despite not being associated with any organization, the menu is the same as above but without the "Org POD" link.

Screen Shot 2022-03-04 at 11 49 57 AM

---

## Other considerations
In the account signup form we currently ask for, but don't require, the user's name. If the user currently logged in did provide a name, we show that in addition to their email address, as shown in the first screenshot in this ticket.

In my mockups above, I haven't shown the user's name. Unless we make name a required field, I'm not sure it makes sense to display it. In theory, we could use logic to display the user's name if it exists, and otherwise display their email address:

Screen Shot 2022-03-04 at 10 53 45 AM

but I'm not sure if it is necessary and I'd personally prefer to just display email addresses, or require name during sign up and display the user's name instead of email address. Maybe someone else has stronger feelings about using names than me, though.

marlo-longley:
Fixes #446 - the UI should cover all the scenarios described by the ticket, worked with @thatbudakguy 

Looking for input on:
- The way we found the user's organizations and if there's a better Rails magic way
- Tests, how to cover the cases of being logged in with different roles (admin, and user who is a member of an organization)

For the UI colors, we kept things in default dark mode for now until future design work specifies any changes.
Screen Shot 2022-03-10 at 16 38 39

### DIFF
--- a/app/views/shared/_nav.html.erb
+++ b/app/views/shared/_nav.html.erb
@@ -11,21 +11,33 @@
       <li class='nav-item'>
         <%= link_to 'Activity', activity_url, class: 'nav-link', data: { turbolinks: false } %>
       </li>
-      <% if can? :manage, User %>
-        <li class='nav-item'>
-          <%= link_to 'Users', site_users_url, class: 'nav-link' %>
-        </li>
-      <% end %>
     </ul>
     </div>
 
     <ul class="navbar-nav ml-auto">
       <% if current_user %>
-        <li class="nav-item">
-          <%= link_to current_user, edit_user_registration_path, class: 'nav-link' %>
-        </li>
-        <li class="nav-item">
-          <%= link_to 'Logout', destroy_user_session_path, class: 'nav-link', data: { method: :delete } %>
+        <li class="nav-item dropdown">
+          <a class="nav-link dropdown-toggle" href="#" id="navbarDropdown" role="button" data-bs-toggle="dropdown" aria-expanded="false">
+            <%= current_user %>
+          </a>
+          <ul class="dropdown-menu dropdown-menu-dark" aria-labelledby="navbarDropdown">
+              <% current_user.organizations.each do | organization | %>
+                <li class="nav-item"><%= link_to "#{organization.name} POD", organization, class: 'nav-link' %></li>
+              <% end %>
+             <li class="nav-item">
+              <%= link_to 'Edit profile', edit_user_registration_path, class: 'nav-link' %>
+            </li>
+            <li><hr class="dropdown-divider"></li>
+            <% if can? :manage, User %>
+              <li class='nav-item'>
+                <%= link_to 'Manage users', site_users_url, class: 'nav-link' %>
+              </li>
+            <% end %>
+            <li><hr class="dropdown-divider"></li>
+            <li class="nav-item">
+              <%= link_to 'Logout', destroy_user_session_path, class: 'nav-link', data: { method: :delete } %>
+            </li>
+          </ul>
         </li>
       <% else %>
         <li class="nav-item">
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -12,6 +12,10 @@ class User < ApplicationRecord
 
   has_many :uploads # rubocop:disable Rails/HasManyOrHasOneDependent
 
+  def organizations(which_roles = %i[owner member])
+    Organization.with_roles(which_roles, self)
+  end
+
   def to_s
     return email if name.blank?
 
--- a/spec/features/nav_dropdown_spec.rb
+++ b/spec/features/nav_dropdown_spec.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe 'Using the dropdown menu in the navbar', type: :feature do
+  let(:organization) { create(:organization, name: 'Best University') }
+  let(:user) { create(:user) }
+
+  before do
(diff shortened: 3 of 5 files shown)


app/views/shared/_nav.html.erb
```diff
@@ -11,21 +11,38 @@
...
+            <%= current_user %>
+          </a>
+          <ul class="dropdown-menu dropdown-menu-dark" aria-labelledby="navbarDropdown">
+            <% current_user.roles.each do |role| %>
+              <% if role.resource_type == "Organization" %>
+                <li class="nav-item">
+                  <% @organization = Organization.find(role.resource_id) %>
+                  <%= link_to "#{@organization.name} POD", @organization, class: 'nav-link' %>
+                </li>
+              <% end %>
```
corylown, line 30: Maybe something like?:

```
current_user.roles.where(resource_type: 'Organization').each do |role|
  <li class="nav-item">
    <%= link_to "#{role.resource.name} POD", role.resource, class: 'nav-link' %>
  </li>
<% end %>
```

As an aside, when creating variables in templates it's usually a good idea to use local variables instead of instance variables (`organization` vs. `@organization`) to keep the scope limited.

corylown:
For testing users with different roles/attributes you'll need to create them using factorybot. You're already using factory bot when you set the `current_user` to `:user`:  `let(:current_user) { create(:user) }`. You probably already found it, but that `:user` is defined here: https://github.com/pod4lib/aggregator/blob/main/spec/factories/users.rb#L4-L9.

You can create a new user or add traits to this user to test various scenarios. Some links to docs about how to define and use factories:

- https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#defining-factories
- https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#using-factories

app/views/shared/_nav.html.erb
cbeer: You can also use [rolify](https://github.com/RolifyCommunity/rolify#6-resource-roles-querying) to get the organizations, e.g.:

```ruby
Organization.with_roles([:owner, :member], current_user)
```

(although logic like that is probably a good candidate for pulling up to the `User` or `Organization` model?)

marlo-longley:
Here are some updates and more tests which may be overkill!

I was curious about the preferred place to test the links in the menu - in the template spec vs. feature spec like we discussed this morning. Right now it's sort of split between `spec/features/nav_dropdown_spec.rb` and the existing login/logout link tests in `spec/views/shared/nav.html.erb_spec.rb`

thatbudakguy (approved):
these tests look great to me! i think moving the organization specific stuff to a combo of method tests for `User.organizations` and a feature test for the nav makes sense.

corylown (approved):

Issue #459, opened 2022-03-09
apply a theme to change bootstrap's default colors


thatbudakguy:
right now the site applies bootstrap "dark" styles in a few places, but not in a broad or careful way, which has generated some UI problems (see #456 and #443).

the "vanilla" bootstrap elements, like the primary color for buttons, also very obviously show that the site is "default bootstrap" and hasn't been customized.

most likely we'd want to override all of [bootstrap's top-level "theme colors"](https://getbootstrap.com/docs/5.1/customize/color/#theme-colors) so that they play well together and have an overall sense of identity. here are the defaults:
```yaml
- name: primary
  hex: "#0d6efd"
- name: secondary
  hex: "#6c757d"
- name: success
  hex: "#28a745"
- name: danger
  hex: "#dc3545"
- name: warning
  hex: "#ffc107"
- name: info
  hex: "#17a2b8"
- name: light
  hex: "#f8f9fa"
- name: dark
  hex: "#343a40"
```

we might also want to make an overarching decision about using the "dark versions" of components, like the dark navbar.

Issue #475, opened 2022-03-14
Do we want to allow anyone to register for a user account?


ggeisler:
Currently anyone can successfully register for a POD account. This might not be what we want?

If we do want to be fairly open to self-registration, do we at least want a POD admin to confirm new accounts before they're activated, to prevent spamming, etc.?

anarchivist:
I would tap @JohnMarkOckerbloom, because I would expect we don't want it to be open registration for the following reasons:

1) In the current context, we want to make sure that IPLC members can easily get access, as the ReShare/BorrowDirect project is our priority
2) Data consumers must abide by the [data provider and usage framework](https://docs.google.com/document/d/1EZG7z_A13jxPfHbYhT7Ae4Ays10fhecqfoGJhqrcjD8/edit). While it's lightweight, it does not mean there are no limitations and the data is fully openly reusable.

ggeisler:
> If admins rather than users initiate account creation, users should probably reset their passwords as soon as possible (since they didn't create them).

If I remember correctly, the current organization invite path is: 

1. The organization admin enters an email address to invite someone (using the "Invite" action)
2. The person at that email address receives an email with a link to create a POD account
3. That person follows the link (to the POD app Sign up page) to create their new account with a password of their own choosing

I think that path will still work, even with the proposed user role revamp. So we wouldn't have to worry about telling the user to reset their password.

The question about this that still remains for me (it has implications for the homepage redesign and possibly some places in the Devise-related pages) is: do we want to initiate all new account creation "top down" (via a POD org admin or a superuser sending someone an invite to create an account, with the admin/superuser finding out about this new user via some means **external** to the POD application), or do we want to have a way for a potential new user visiting the POD application to request an account on their own, via a link in a public page of the application?

anarchivist:
Part of the question to me is what action does requesting account lead to in the aggregator? A couple options:

* Creating a "stub" user that can be vetted (but then requires us to notify admins and provide a clear way to triage requests)
* A feedback or "request an account" form, which requires manual action on the part of the admins.
* Additional pointers in the documentation/wiki?

ggeisler:
> Creating a "stub" user that can be vetted (but then requires us to notify admins and provide a clear way to triage requests)
> A feedback or "request an account" form, which requires manual action on the part of the admins.

[...]

Issue #266, opened 2020-12-03
Error when attempting to sign up


JustinUhr:
When trying to sign up, after entering email and password in both fields and clicking 'Sign Up', I got the following error:
![POD_signup_error](https://user-images.githubusercontent.com/56878072/101081915-b2614980-3578-11eb-9dbe-8ef70e36596d.png)

I was able to create an account successfully after being invited.
(sorry for the dark theme in the screenshot)

thatbudakguy:
i think given discussion on #475 this can be closed, since open signups should (soon) be impossible — it will be done by admin invite instead. that will also remove the need for the captcha.

corylown:
Individual sign-ups are going away so I'm closing this issue.

PR #484, opened 2022-03-16
Update design for Devise pages

### LINKED ISSUE #436: Update layout of Login/User profile page (Devise layout)
Related to #444.

The pages related to the user account tasks (registration, login, user profile) are unnecessarily narrow, creating a poor experience when using a browser with a narrow viewport:

Screen Shot 2022-02-28 at 3 01 39 PM

---

This appears to happen because the [Devise layout file](https://github.com/pod4lib/aggregator/blob/c4552721cf8b4e9532b36b5f359cf991cb8254e9/app/views/layouts/devise.html.erb) includes the CSS class `w-25`, which tells that div to use only 25% of the container width. 25% width is sufficient when the viewport is a decent desktop size (though even there wider than 25% would be perfectly fine), but not at narrower viewports.

I'd suggest replacing the `w-25` class with some Bootstrap column classes, which aren't currently used in this layout file. ~For example using `col-md-6` might work for this div, using half the page width at `md` and larger, while using the full width on mobile browsers. That approach might require adding a `row` class somewhere above, not sure off-hand.~ See #444 for a more thought-through suggestion.

If we want to use the simplest possible approach, I think just removing the `w-25` would be okay (form inputs would use the full width of the page, which is unnecessary at `md` and wider, but is much better than being too narrow), but doing #444 instead of this ticket would result in a much better overall user experience.

### LINKED ISSUE #444: Update layout and content of Devise-based pages
In issue #436 I suggested a simple fix to the Devise-based sign-up/login pages. That issue only addresses the current layout problem with those pages, though, and as I work through the app more I've noticed other issues with the sign-up/login pages that ideally would also be addressed to create a more attractive and clear user experience. If we don't have time to revamp these pages as suggested below, #436 is the minimum we should do, but it would be better to do the more thorough revamp suggested below.

## Basic layout changes
As I noted in #436, I think we should introduce Bootstrap columns into these pages for better layout. In the mockups below, I've used a wrapper div with `class="col-md-8 offset-md-2` to position the form in the middle of the page, at the `md` breakpoint and larger, and keeping the width of the forms plenty wide but not full-page width. At below `md` I think it will be fine to give the form full-page width.

Other consistent layout choices:
- Center the page title
- Add a horizontal rule/bottom border to separate the main form from the potentially relevant but secondary ("Sign up", "Log in") and tertiary ("Forgot your password?") link actions
- Left-align the secondary actions for a given page with the left-edge of the form
- Right-align the tertiary actions with the right-edge of the form

Note that some page titles, field labels, and button labels are the same as the Devise default but others are updated a bit for better clarity.

Here's how the various pages would more or less look after those changes are made:

## Set password
This is the form you get if you are emailed an invite to sign up with POD.

set-password

---

## Sign up

sign-up

---

## Log in

log-in

---

## Forgot password

forgot-password

---

## Resend confirmation instructions

resend

---

marlo-longley:
Fixes #444 and by extension fixes #436 @ggeisler 

Aside from HTML and CSS changes, this PR adds a new entry to `locales/en.yml` to override the default label for email addresses.

Notes:
- The button colors, which are tied to the bootstrap theme, I've left as default/primary, so for now they don't match Gary's mockups. But this should become aligned after addressing  #459

Screen Shot 2022-03-16 at 16 39 23
Screen Shot 2022-03-16 at 16 38 54
Screen Shot 2022-03-16 at 16 39 30

### DIFF
--- a/app/views/layouts/devise.html.erb
+++ b/app/views/layouts/devise.html.erb
@@ -2,7 +2,7 @@
   <%= content_for(:header) %>
 
   <div class="container">
-    <div class="col-md-8 offset-md-2 devise-wrap">
+    <div class="col-md-6 offset-md-3 devise-wrap">
       <%= yield %>
     </div>
   </div>
--- a/app/views/layouts/devise.html.erb
+++ b/app/views/layouts/devise.html.erb
@@ -2,7 +2,7 @@
   <%= content_for(:header) %>
 
   <div class="container">
-    <div class="mr-auto ml-auto w-25 mt-5 mb-5">
+    <div class="col-md-8 offset-md-2 devise-wrap">
       <%= yield %>
     </div>
   </div>
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -123,3 +123,45 @@ input[type=file][data-direct-upload-url][disabled] {
 .bg-local-practice {
   background-color: lighten($warning, 40%) !important;
 }
+
+.devise-wrap {
+  padding-top: $spacer;
+
+  h1 {
+    text-align: center;
+    padding: $spacer 0;
+  }
+
+  // the main form elements
+  form > .form-group {
+    padding-bottom: $spacer * 2;
+  }
+
+  form > .form-group:last-child {
+    border-bottom: 2px #000 solid;
+  }
+
+  form > .form-group > label {
+    padding-bottom: $spacer / 2;
+  }
+
+  .form-check {
+    margin-bottom: 0;
+    padding-left: 0;
+  }
+
+  .form-check .form-check-input {
+    // override negative margin-left from BS
+    margin-left: 0;
+    margin-right: 5px;
+  }
+
+  // additional links below main form
+  > .links-list > li {
+    padding-top: $spacer;
+  }
+
+  > .links-list > li > a {
+    text-decoration: none;
+  }
+}
--- a/app/views/devise/shared/_links.html.erb
+++ b/app/views/devise/shared/_links.html.erb
@@ -1,27 +1,27 @@
-<div class="form-group">
+<ul class="list-unstyled links-list">
   <%- if controller_name != 'sessions' %>
-    <%= link_to t(".sign_in"), new_session_path(resource_name) %><br />
+    <li><%= link_to t(".sign_in"), new_session_path(resource_name) %></li>
   <% end -%>
 
   <%- if devise_mapping.registerable? && controller_name != 'registrations' %>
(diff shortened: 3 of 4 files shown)


ggeisler:
Looks good to me @marlo-longley (but will defer to a developer for formal review.

app/views/layouts/devise.html.erb
```diff
@@ -2,7 +2,7 @@
   <%= content_for(:header) %>
 
   <div class="container">
-    <div class="mr-auto ml-auto w-25 mt-5 mb-5">
+    <div class="col-md-8 offset-md-2 devise-wrap">
```
thatbudakguy, line 5: the fields look super long to me at this size. what if we did `col-md-6 offset-md-3`?

thatbudakguy (approved):
everything looks fine to me, except that the text fields seem super wide on desktop.

app/views/layouts/devise.html.erb
ggeisler: The original design included links both left- and right-aligned below the horizontal rule, and if I remember correct in some cases (e.g., the Forgot your password? page) going smaller than `col-md-8` put the adjacent links too close together to look good. And there's no harm in an overly wide input field on this page so that seemed the best compromise.

I guess if we're going to stick to left-aligned only links, the above situation is no longer in play, so sure, I guess we could go with `col-md-6 offset-md-3` as long as that still works at all breakpoints.

marlo-longley: Oops, when I commented my view hadn't updated since Gary's comment.

thatbudakguy: @ggeisler thx for context. i think let's just commit to keeping everything left-aligned so we can shrink the overall container a bit on large screens.

marlo-longley:
@thatbudakguy @ggeisler here's the new look; pushed the commit.
Old
Screen Shot 2022-03-16 at 16 39 30
New
Screen Shot 2022-03-17 at 11 28 02

ggeisler:
@marlo-longley Since this hasn't been merged yet, note that the decision made in #475 that there be no direct user registration means we should remove the "Sign up" link from all user account pages. 

Not sure if you'd prefer to do that via another commit to this PR or if you want to do it as a separate issue after this is merged.

thatbudakguy:
@ggeisler I took a look at this and I think it'd be better to do it as a separate issue/PR, since I think we will also need to remove any language that appears when the user gets to the sign up page via a different flow:

accesses login-only page -> prompted to sign in or sign up -> clicks sign up

there are some other related things to remove on the backend too, like the reCaptcha and form for signing up (which have some existing problems as noted in #266).

thatbudakguy:
new issue for restricting signup is #499.

PR #485, opened 2022-03-16
Add more dev environment info to README

marlo-longley:


thatbudakguy (approved):
thanks for updating this!

PR #497, opened 2022-03-21
Tweak the user nav dropdown

### LINKED ISSUE #496: Adjust details of user menu in topbar
Several issues of varying importance I noticed:

- If you aren't an admin, we show an extra menu divider (which we do need when you are an admin):

Screen Shot 2022-03-21 at 11 49 10 AM

- I forget if this is adjustable via a Bootstrap class or not, but it feels odd to me that the menu items are aligned on the opposite side from the icon you use to display the user menu. In other words, is it possible to right-align the menu so that it drops down directly below the icon?

Screen Shot 2022-03-21 at 11 47 49 AM

- The dark background of the menu with the gray text: not sure whether it is accessible contrast or not (I'll look at accessibility details later) but visually it looks muddy to me. I think a white background menu would work better. Since we do have plans to reevaluate the application color palette separately, we could wait on this, but I'm pretty sure a light background menu will work with whatever do change the topbar background color to.

(For the last two items, look at Github's equivalent user menu in the topbar, which basically does what I'm suggesting for both of those.)

- Also, maybe this needs to be a product owner decision (@JohnMarkOckerbloom), but since we don't currently require a registered user to have provided a name, I'd suggest we only display the email address as the user menu link (as in the first screenshot in this comment). That would take up less space in the topbar while providing more consistency between users with provided names and users without, and I personally don't see the need to show the user's actual name versus their email address.

cbeer:
Screen Shot 2022-03-21 at 12 38 05

Fixes #496

### DIFF
--- a/app/views/shared/_nav.html.erb
+++ b/app/views/shared/_nav.html.erb
@@ -20,7 +20,7 @@
           <a class="nav-link dropdown-toggle" href="#" id="navbarDropdown" role="button" data-bs-toggle="dropdown" aria-expanded="false">
             <%= current_user.email %>
           </a>
-          <ul class="dropdown-menu dropdown-menu-dark" aria-labelledby="navbarDropdown">
+          <ul class="dropdown-menu dropdown-menu-dark dropdown-menu-end" aria-labelledby="navbarDropdown">
               <% current_user.organizations.each do | organization | %>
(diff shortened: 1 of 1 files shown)


ggeisler:
👍

thatbudakguy (approved):

Issue #454, opened 2022-03-09
Add seed data for development


corylown:
Currently, the seed data is pulled from pod prod and maybe this is not practical and we want something different for development work.

PR #498, opened 2022-03-22
Return additional file metadata with stream JSON response.

corylown:
Part of #454. I'd like to know the file size in the stream/file JSON response in order to filter out large files from the seed data. This adds some metadata to the JSON response that's available in the ResourceSync XML response.

### DIFF
--- a/app/views/streams/show.json.jbuilder
+++ b/app/views/streams/show.json.jbuilder
@@ -10,7 +10,7 @@ json.uploads @stream.uploads do |upload|
     json.id file.id
     json.filename file.filename
     json.hash "md5:#{Base64.decode64(file.checksum).unpack1('H*')}"
-    json.type file.content_type
+    json.content_type file.content_type
     json.length file.byte_size
     json.download_url download_url(file)
   end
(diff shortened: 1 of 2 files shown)


app/views/streams/show.json.jbuilder
```diff
@@ -9,6 +9,9 @@ json.uploads @stream.uploads do |upload|
   json.files upload.files do |file|
     json.id file.id
     json.filename file.filename
+    json.hash "md5:#{Base64.decode64(file.checksum).unpack1('H*')}"
+    json.type file.content_type
```
cbeer, line 13:
```suggestion
    json.content_type file.content_type
```

🤷‍♂️ . I'd prefer to keep consistent with the ruby-land method, but `type` feels especially generic (and likely to conflict with JSON-LD or JSON:API if we ever go there..).

corylown: Yeah -- that makes sense. I was torn about being consistent with the ResourceSync response or being more, uh, descriptive.

corylown: I think I'll switch `length` to `byte_size` too? so I don't lose my mind

cbeer (approved):

PR #504, opened 2022-03-23
Refactor seed data generation

### LINKED ISSUE #454: Add seed data for development

corylown:
Definitely open to suggestion on this, but the proposal:

- Simplify the regular `rake db:seed` task to create one institution and load small sample files that are included in the repository. This seems most useful for a developer just getting started and wanting some basic seed data.
- Move the prior seed script to a rake task `rake agg:seed_from_api`. This remains configurable via `settings.yml`. I opted to keep this script available in case it's something we want to use to populate staging from prod or other uses.

Closes #454


marlo-longley (approved):
@corylown I tried out this branch (ran `Organizations.destroy_all` to clear my DB) -- and it worked great.

PR #508, opened 2022-03-24
Use Font Awesome SVG icons to indicate file status

### LINKED ISSUE #425: Clarify meaning of status icons in file list
In the organization view (for example: https://pod.stanford.edu/organizations/stanford) there is a table of contributed files that displays a status indicator:
Screen Shot 2022-02-03 at 9 58 02 AM

It's unclear what these status symbols mean. It would be helpful to revise the design to provide a tool tip or legend. It may also be worth reconsidering the overall design of how these states are indicated.

The possible states are:

- File is valid MARC (`:success`  ✅)
- File of deletes (`:deletes` 🗑)
- Invalid MARC file (`:invalid` ❌ )
- File contains something other than MARC or deletes (`:not_marc` ❓)
- Unknown (`:unknown`🔄)

marlo-longley:
Fixes #425 

There were many ways to approach changing the icons (css only vs. markup output vs. rails helpers) but this approach avoided the need to import a font awesome or bootstrap icons library, and stayed in line with the existing logic. 

I also wanted to test for the fact the the correct status classes were being applied - there may be better places to test as well. 

Here's how it's looking:
Screen Shot 2022-03-24 at 14 20 31

(unknown status:)
Screen Shot 2022-03-24 at 14 25 10

Other styling issues such as vertical alignment within the table is outside the scope of this issue and can be fixed when we work on tables in the larger redesign.


app/helpers/bootstrap_icons_helper.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+# Helper to render Bootstrap Icons
+module BootstrapIconsHelper
+  # refer https://icons.getbootstrap.com/
+  def icon(icon, options = {})
```
corylown, line 6: I think I'd move this method to the existing `ApplicationHelper` and maybe name the method `bootstrap_icon` for clarity. Maybe add a simple test too?

app/helpers/bootstrap_icons_helper.rb
```diff
@@ -0,0 +1,15 @@
...
+    file = File.read("node_modules/bootstrap-icons/icons/#{icon}.svg")
+    content_tag(:span, class: options[:class].presence) do
```
corylown, line 8: As far as I can tell `options[:class]` and `options[:class].presence` both just return nil if there's not a `:class` key so I don't think you need to call `.presence`

config/settings.yml
```diff
@@ -19,3 +19,11 @@ marc_fixture_seeds:
     - Cornell
     - Princeton
     - Stanford
+
+# refer to https://icons.getbootstrap.com/ for icon class names
+metadata_status_icons:
```
corylown, line 24: I'm wondering why this is in settings? It doesn't seem like something that we're likely to configure differently per environment. Maybe a helper method would work?

in `ApplicationHelper`:
```ruby
def metadata_status_icons(status)
  case status
  when :deletes then 'file-earmark-excel-fill'
  when :invalid then 'exclamation-circle-fill'
  # etc...
end
```

corylown (changes requested):
A few inline questions/comments. The original issue shows a legend above the table explaining the meaning of the icons, which I think would be nice to add to this PR: https://github.com/pod4lib/aggregator/issues/425

marlo-longley:
@corylown thanks for your careful review, I will address.

marlo-longley:
@corylown I've addressed your comments, thanks for your help!
You can see the legend looks like this now:

Screen Shot 2022-03-30 at 17 42 18

corylown (approved):

PR #513, opened 2022-03-28
Remove incorrect info from README

marlo-longley:


thatbudakguy:
could we also add info about what @corylown mentioned this morning? i.e. adding

```rb
  config.active_job.queue_adapter = :sidekiq
```

to `development.rb`

Issue #516, opened 2022-03-29
Should uploads be surfaced as resources to users?


thatbudakguy:
at the moment, we don't know the reason why Uploads (as a resource) exist. so far, we've assumed that there's a technical reason why it made sense to allow files to be grouped at upload time and to have that grouping persist as an independently named entity that links Files to Streams.

this is reflected in some inconsistencies in the UI, including the [unstyled upload list page](https://pod.stanford.edu/organizations/stanford/uploads/), [upload show page](https://pod.stanford.edu/organizations/stanford/uploads/19285), and ["activity" page](https://pod.stanford.edu/dashboard/uploads) as noted in https://github.com/pod4lib/aggregator/issues/479. the job status tracker also refers to uploads in some cases, notably the `Extract Marc Record Metadata Job`.

organizations seem to be treating the concept of uploads differently: some of them always upload a single file per upload and name the upload after that file (e.g. duke), while some upload multiple files and may not specify a name. dartmouth seems to take advantage of the ability to point at a file already uploaded elsewhere on the web and [doesn't "upload" files at all](https://pod.stanford.edu/organizations/dartmouth/uploads/8594).

presuming that there's a compelling technical reason for Uploads to exist as a resource, this issue asks whether that design decision needs to be surfaced to users (and if so, how).

a potential user story from the "no" perspective:

_As a user, I want the concept of uploads to be hidden from me, so that I don’t have to think about how files were grouped when they were uploaded._

JohnMarkOckerbloom:
Thanks for opening this issue!  I suspect my confusion about file and upload names in today's standup may also be present in others.  I'd actually assumed that mentions of "uploads" in the stats page referred to uploaded files, rather than to sets of files organized into an "upload" grouping.  And while our documentation talks about files and the streams that contain them, there isn't really anything discussing uploads as distinctive data structures.  Nor have I yet seen a compelling reason to maintain them as distinctive data structures.  (I have seen some hints that files grouped into an upload are processed together in ways that files that are not aren't, but it's unclear that that's actually happening, what specifically happens, and whether directing files to be processed together is necessary.)

In the absence of a compelling reason, I'm inclined to apply the "do not multiply entities unnecessarily" principle, and simplify the provider data model and just report on files and streams.  But I'm open to justifications for keeping "uploads" as entities as well.

PR #518, opened 2022-03-30
Paginate upload lists on stream and org pages

### LINKED ISSUE #450: Organization pages with many uploaded files are slow to load.

thatbudakguy:
Fixes #450


corylown (commented):
One thing that's a little strange is that we're paging on uploads but effectively showing files in the UI -- this means that an upload of 50 files is counted as one item by Kaminari and leads to some pretty strange paging behavior because sometimes you see many files per page and sometimes few depending on the number of files per upload. Maybe we should page on files instead, but I'm not sure if we lose some of the metadata we're displaying then.

One other small thing -- could use some padding between the table and paging controls:
Screen Shot 2022-03-30 at 3 42 38 PM

thatbudakguy:
> we're paging on uploads but effectively showing files in the UI

good point. this is maybe more evidence for #516 that it makes more sense to hide uploads or do away with them entirely?

corylown (approved):

PR #521, opened 2022-03-30
Clarify labels for change password form fields

### LINKED ISSUE #156: Cannot change my password
The User edit form gives errors on save:

```
2 errors prohibited this user from being saved:

    Password confirmation doesn't match Password
    Current password is invalid
```

Even though I entered my real password and the confirmation values matched (I tried multiple times and even using the auto-generated passwords on my browser.)

Screen Shot 2020-11-11 at 2 28 06 PM

thatbudakguy:
Fixes #156

### DIFF
--- a/app/views/devise/registrations/edit.html.erb
+++ b/app/views/devise/registrations/edit.html.erb
@@ -13,9 +13,9 @@
 
   <%= f.text_field :title, label: 'Job title' %>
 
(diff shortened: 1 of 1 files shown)


thatbudakguy:
@marlo-longley thanks for the reminder about the locales file, I forgot about it! weirdly though it looks like devise doesn't use it for this ["change password" view](https://github.com/heartcombo/devise/blob/main/app/views/devise/registrations/edit.html.erb)... kinda seems like they should? but for the time being since they don't, I guess we don't need to either. idk

corylown (approved):

PR #525, opened 2022-03-31
Do not expose inter-stream delete deltas via UI or API; closes #522

### LINKED ISSUE #522: Remove inter-stream deletes from resource lists and the normalized files list.
Decided in meeting on 3/30/22 with @anarchivist & @JohnMarkOckerbloom that we're going to remove the inter-stream delete deltas from the resource lists.

We will leave the existing `removed_since_previous_stream` method in place but will not expose the generated file via the UI.

Places where this needs to be removed:

- https://github.com/pod4lib/aggregator/blob/main/app/views/streams/show.xml.builder#L12-L14
- https://github.com/pod4lib/aggregator/blob/main/app/views/streams/normalized_dump.xml.builder#L14-L16
- https://github.com/pod4lib/aggregator/blob/main/app/views/streams/_normalized_dump.html.erb#L13

corylown:


thatbudakguy:
just saw on #525 that we're keeping the methods around. could we add a comment at least so that whoever finds this next knows that the methods are deprecated/not used (or even mark them as deprecated?) I feel like we've spent so much time already on this project trying to interpret old code and I don't want to leave it any worse than we found it

PR #526, opened 2022-03-31
Redesign edit profile page

### LINKED ISSUE #447: Update layout of user profile page
Related to #436 and #444.

As noted in #436, the Devise-based user pages include the `w-25` Bootstrap class, which makes the main content area of these pages unnecessarily narrow in most viewport widths:

Screen Shot 2022-03-04 at 1 48 25 PM

Removing the `w-25` class should help the user profile page as well, however, if we can tackle the suggested approach in #444, I'd suggest we could also make some other minor improvements to the user profile page, as outlined below.

Below is a mockup of the suggested improvement, followed by notes on specifics:

Screen Shot 2022-03-04 at 1 49 47 PM

---

- **A:** The suggested layout assumes the work in #444 has been completed, so the centered form and page title would likely be inherited from that work. (This layout doesn't match pages elsewhere in POD, but I think the user-related pages are distinct enough that it's fine, maybe even desirable, for them to have a separate layout.)

- **B:** Page title is renamed `User profile`. This should also be changed to an `h1` from the current `h2`, since the application is using page titles as `h1`s elsewhere. (I hope to do a more thorough a11y analysis of things during the workcycle.)
- **C:** This is likely the trickiest part of this issue: separating the basic user profile fields from the change password fields. I think it's important, though, because the current form is potentially confusing (e.g., do you need to enter your password in the first password field if you are just changing your job title?). Also, I believe currently **you are** required to provide your current password to change the basic profile fields, which might be a native Devise thing but seems unnecessary for how the POD application will be used. This page is not accessible unless the user is logged into POD, which seems sufficient proof of identity to me for the user to change their name or job title (I wonder if the email address input should be a read-only field, though. Perhaps we'd want to prevent a user who was invited to join an organization using a specific email address from later changing that email address without the person who invited them to their organization knowing?). So my suggestion is those three fields make up a separate form that the "Update" button (`btn-primary`) submits. 
	- The label of the first field is changed to "Email address". The input might be disabled, if people agree with my concern above.
	- "Cancel" returns the user to the previous page, same as the "Back" action does on the current profile page.
- **D:** This section splits out the password fields as a new form for the user to change their password. 
	- Section heading "Change password" is an `H2`
	- Fields are re-ordered so the user starts by entering their current password.
	- Hint is removed from current password field since the distinct section makes it pretty clear you need to enter all fields to change your password.
	- Field labels are changed for "New password" and "Confirm new password"
	- Submit button label is "Change password" and is a `btn-secondary`
- **E:** Change the section heading to "Cancel your account" and make it an `h2`. Style the "Cancel account" button with `btn-danger` so it is more clearly a destructive action.

## Additional (future?) layout updates
This might be too much for the current workcycle but in the future, the sections on this page might benefit from being separated  into three tabs:

- Edit profile
- Change password 
- Cancel account

For example, below I've kept the same section content as above (except for adding an intro paragraph to the Cancel account tab), but split the sections into tabs:

Screen Shot 2022-03-04 at 2 21 01 PM

---

Screen Shot 2022-03-04 at 2 21 12 PM

---

Screen Shot 2022-03-04 at 2 21 25 PM

---

This arrangement would make things less cluttered and provide a better foundation for adding new user-specific features in the future (just add a new tab, or add fields to a current tabbed section) without creating an long, unwieldy page.

marlo-longley:
Fixes #447 

The main change of this PR is allowing users to edit non-password items (i.e. name, email) without entering their current password. 

@ggeisler The spacing between form items is slightly taller than your design indicated but close. The helper gem that provides `bootstrap_form_for` is adding extra stuff into the form template that conflicts with CSS I wrote previously for the Devise forms. I personally don't like this gem and would be down to rip out the dependency, but I think refactoring the forms is a separate low-priority issue. I can definitely do a workaround for the spacing however, if you think it's not acceptable.

Screen Shot 2022-03-31 at 15 54 29

I appreciate your analysis in the issue and agree that for now, a single vertical layout is easiest, rather than implementing tabs. Doing tabs introduced complications with routing.

### DIFF
--- a/spec/features/edit_profile_spec.rb
+++ b/spec/features/edit_profile_spec.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe 'editing your user profile', type: :feature do
+  context 'with a user' do
+    let(:user) { create(:user, email: 'test@stanford.edu') }
+
+    before do
+      login_as(user, scope: :user)
+    end
+
+    it 'requires current password to change password' do
+      visit '/users/edit'
+      fill_in 'user_password', with: '123'
+      click_on 'Change Password'
+      expect(page).to have_content "Current password can't be blank"
+    end
+
+    it 'allows non-password changes without passwords' do
+      visit '/users/edit'
+      fill_in 'user_name', with: 'Nice Name'
+      click_on 'Update'
+      # successful update redirects to root page
+      expect(page).to have_content 'Your account has been updated successfully'
+    end
+  end
+end
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -137,15 +137,16 @@ input[type=file][data-direct-upload-url][disabled] {
   }
 
   // the main form elements
-  form > .form-group {
-    padding-bottom: $spacer * 2;
+  .form-group {
+    padding-top: $spacer * 2;
   }
 
-  form > .form-group:last-child {
+  .form-group:last-child {
+    padding-bottom: $spacer * 2;
     border-bottom: 2px #000 solid;
   }
 
-  form > .form-group > label {
+  .form-group > label {
     padding-bottom: $spacer / 2;
   }
 
@@ -167,5 +168,20 @@ input[type=file][data-direct-upload-url][disabled] {
 
   > .links-list > li > a {
     text-decoration: none;
+  }
+}
+
+.user-profile-section {
+  margin-top: $spacer * 2;
+
+  .actions > span {
+    margin-right: $spacer * 2;
+    a {
+      text-decoration: none;
+    }
+  }
+
+  #cancel_form {
+    margin-top: $spacer * 2;
   }
 }
--- a/app/controllers/registrations_controller.rb
+++ b/app/controllers/registrations_controller.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+# Subclass of devise controller so we can control edit profile logic
+class RegistrationsController < Devise::RegistrationsController
+  protected
+
+  def update_resource(resource, params)
+    # Require current password if user is trying to change password.
+    return super if params['password']&.present?
+
+    # Allow user to update other registration info without password
+    resource.update_without_password(params.except('current_password'))
+  end
+end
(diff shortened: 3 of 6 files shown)


ggeisler:
@marlo-longley Sure, I think the biggest priority was to separate the first three fields from the change password related fields, because that part is just confusing no matter what you are trying to do. So this fixes that.

I'd still like to see a tabbed UI and better spacing (e.g., there appears to be more vertical spacing below the "Change password" heading than the "Cancel account" heading) but I agree that's lower-priority given all of the other things we're trying to update in the application now.

marlo-longley:
Thanks @ggeisler - I just fixed the vertical spacing issue for the Cancel button that you caught. Thanks!

spec/features/edit_profile_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+      click_on 'Change Password'
+      expect(page).to have_content "Current password can't be blank"
+    end
+
+    it 'allows non-password changes without passwords' do
+      visit '/users/edit'
+      fill_in 'user_name', with: 'Nice Name'
+      click_on 'Update'
+      # successful update redirects to root page
+      expect(page).to have_current_path('/')
```
corylown, line 25: Maybe this?

```suggestion
      expect(page).to have_content 'Your account has been updated successfully.'
```

spec/features/edit_profile_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    before do
+      login_as(user, scope: :user)
+    end
+
+    it 'requires current password to change password' do
+      visit '/users/edit'
+      fill_in 'user_password', with: '123'
+      click_on 'Change Password'
+      expect(page).to have_content "Current password can't be blank"
```
corylown, line 17: Single quotes for consistency:

```suggestion
      expect(page).to have_content 'Current password can't be blank'
```

corylown (changes requested):
A couple of inline suggestions. Gary's mockups use different language for the new password and password confirmation fields and I'm wondering if we want to make that change in the PR:
Screen Shot 2022-04-01 at 5 04 58 PM

marlo-longley:
@corylown thanks! The labels for password fields were addressed in a PR by Nick yesterday, so that's why there was a merge conflict. I rebased; should be resolved. I also changed the test as you suggested.

corylown (changes requested):
@marlo-longley sorry! one more thing I just noticed:

This file `spec/routing/registrations_routing.spec.rb` should be named `spec/routing/registrations_routing_spec.rb`

corylown (approved):

PR #536, opened 2022-04-01
Revise job tracker design

### LINKED ISSUE #317: Revisit job trackers design
During the MVP workcycle, we added job trackers to the dashboard so users could get insight into the status of their data as it gets processed. The current display, as reported by staff at Dartmouth, is leading to some confusion:

Screen Shot 2021-02-23 at 2 21 18 PM

We should probably revisit the design and intent for the job trackers, and think about how to accurately and succinctly convey whether files have been successfully processed.

corylown:
This refactors the job status display (and some of the underlying models) to get us closer to @ggeisler 's designs on https://github.com/pod4lib/aggregator/issues/317. I've left out displaying completed jobs for now, although it could be added later if we want. The accordions always display so that the 0 counts are visible to provide feedback that everything is ok and no jobs are processing or stuck.

Various possible display scenarios:
Screen Shot 2022-04-01 at 2 47 30 PM
Screen Shot 2022-04-01 at 2 47 16 PM
Screen Shot 2022-04-01 at 2 47 08 PM

Closes #317

### DIFF
--- a/app/assets/stylesheets/pod-icons.scss
+++ b/app/assets/stylesheets/pod-icons.scss
@@ -1,9 +1,23 @@
 // Access bootstrap color utility variables
 @import 'bootstrap';
 
-.pod-icon > svg, .pod-metadata-status > svg {
+.pod-icon > svg, .pod-metadata-status > svg, .pod-job-tracker-status > svg {
   height: 25px;
   width: 25px;
+}
+
+.pod-job-tracker-status {
+  &.active {
+    color: $info
+  }
+
+  &.retry {
+    color: $warning
+  }
+
+  &.dead {
+    color: $danger
+  }
 }
 
 .pod-metadata-status {
--- a/app/assets/stylesheets/pod-icons.scss
+++ b/app/assets/stylesheets/pod-icons.scss
@@ -1,7 +1,7 @@
 // Access bootstrap color utility variables
 @import 'bootstrap';
 
-.pod-icon > svg, .pod-metadata-status > svg, .pod-job-tracker-status > svg {
+.pod-icon > svg, .pod-metadata-status > svg {
   height: 25px;
   width: 25px;
 }
--- a/app/models/job_tracker.rb
+++ b/app/models/job_tracker.rb
@@ -5,10 +5,6 @@ class JobTracker < ApplicationRecord
   belongs_to :reports_on, polymorphic: true
   belongs_to :resource, polymorphic: true
 
-  def label
-    "[#{job_class.titleize}] #{resource_label}"
-  end
-
   def resource_label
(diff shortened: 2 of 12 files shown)


ggeisler:
Looks good @corylown. I like the link for how to get help.

This is minor but if you want to make a small update, I'd  center the `td`s in the "Status" column (so icons are centered) and right-align the "Completed records" column (`text-end` class). 

I'm slightly worried the blue background of the accordion headers along with the colored status boxes will make the page look sort of visually busy, but that's a minor visual design concern and I'm happy to assess that later and make a minor visual design PR if I think it is worth addressing.

corylown:
Thanks @ggeisler. I've updated the table column alignment. I agree with you about the blue background, but I'd prefer to handle it as a separate PR if want to change it
Screen Shot 2022-04-01 at 4 12 42 PM
.

app/assets/stylesheets/pod-icons.scss
```diff
@@ -1,11 +1,25 @@
 // Access bootstrap color utility variables
 @import 'bootstrap';
 
-.pod-icon > svg, .pod-metadata-status > svg {
+.pod-icon > svg, .pod-metadata-status > svg, .pod-job-tracker-status > svg {
```
cbeer, line 4: Should we just use `pod-icon` class in the HTML?

app/models/job_tracker.rb
```diff
@@ -44,15 +55,17 @@ def total?
...
-  end
-
   private
 
   def number_with_delimiter(*args)
     ActiveSupport::NumberHelper.number_to_delimited(*args)
   end
+
+  def in_sidekiq_set?(set)
+    set.new.map(&:item).any? { |j| j['jid'] == provider_job_id }
```
cbeer, line 65:
```suggestion
    set.new.find_job(provider_job_id)
```

ought to be equivalent, although sidekiq warns:

>  # This is a slow, inefficient operation.  Do not use under normal conditions.
> https://github.com/mperham/sidekiq/blob/main/lib/sidekiq/api.rb#L279-L286

(although I'm pretty sure that warning applies to this implementation too). I'm not sure if there's an even better approach?

corylown: It seems to work ok with a few hundred jobs in the set. Maybe we add a guard clause to the method where if the job set is larger than some number (500?) it punts and just displays everything in the active queue? I dunno. I don't see another way of getting organization specific feedback about their jobs statuses.

corylown: The only other thought I had was to group jobs by age as some indicator that something could be wrong if the job has been sitting around for more than a few days or a week.

cbeer: ok. I guess we can see what happens 🤷‍♂️

PR #543, opened 2022-04-04
Add boolean provider field to organization. Defaults to true.

### LINKED ISSUE #482: Add an attribute to Organization to indicate if an org is consuming data only.
I'm thinking of IndexData specifically, but generally, we may need a way for an organization that is not contributing data, but is consuming data to manage tokens, users, and contact info.

corylown:
We're likely to want some scopes on the model and may need to manage access to types of tokens, but waiting until we know more.

@thatbudakguy ended up doing it your way 🤷 

Closes #482

### DIFF
--- a/app/controllers/organizations_controller.rb
+++ b/app/controllers/organizations_controller.rb
@@ -73,7 +73,7 @@ class OrganizationsController < ApplicationController
   def organization_params
     params.require(:organization)
           .permit(
-            :name, :slug, :icon, :code,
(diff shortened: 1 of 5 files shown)


thatbudakguy (approved):
looks good to me — agree on the choice to keep this small, and handle the implications in following PRs

corylown:
The other way to do this might be to use single table inheritance (https://api.rubyonrails.org/classes/ActiveRecord/Inheritance.html), but maybe this is good enough for now.

Issue #532, opened 2022-04-01
Dockerize local environment - research


ktamaral:
As a developer, I would like to have a containerized local development environment for a quick and repeatable installation and setup process. Creating a dockerized local environment will take some upfront investment for research and implementation, but will reduce onboarding time for new developers and allow for the ability to make changes to the installation consistently across all environments.

Proposal:

- Research and evaluate the complexity of dockerizing the local environment
- The Harvard team has knowledge and experience with dockerizing RoR apps
  - Note the caveat that we aren’t yet sure of the complexity given no prior experience with POD
- The purpose of this proposal is to research the level of complexity to provide more information on whether to proceed with the work

Requirements:
- Dockerization of the application
- Dockerization of other components such as redis instance
- Scripts to populate test data in the local environment
- A docker compose file to orchestrate provisioning of the app and other components
- Instructions for setting up configuration and running the docker compose
- Docker must be optional, any changes made must not affect the manual setup process

thatbudakguy:
I'd be curious to know what the benefits of running the application itself in a container are, given that the deployment environment is not containerized. In my experience, docker is usually used to achieve parity between environments, so having a dev setup that's different from staging/prod in this way doesn't seem optimal.

Re: running the dependent services in containers, I've been using this simple compose file:
```yml
services:
  db:
    image: postgres
    environment:
      POSTGRES_USER: pod4lib
      POSTGRES_PASSWORD: pod4lib
      POSTGRES_DB: aggregator
    ports:
      - "5432:5432"
    volumes:
      - db:/var/lib/postgresql/data

  cache:
    image: redis
    ports:
      - "6379:6379"

volumes:
  db:
```

following the instructions in the README to set environment variables like:
```sh
export REDIS_HOST=localhost
export REDIS_PORT=6379
export DATABASE_URL=postgresql://pod4lib:pod4lib@localhost:5432/aggregator
bundle exec rails server  # or other commands
```
has worked for me.

thatbudakguy:
also in case it's helpful, here's [a `Dockerfile` that we used on a previous project](https://github.com/sul-dlss/dlme/blob/eaf8be093db5cbdc41ba833f16733c8417e52667/docker/Dockerfile) where we had a rails-based app with a worker/sidekiq instance, and wanted to build production and dev versions of both of those services. we used build targets to keep the dependencies separate for each service version.

ktamaral:
Pull Request: https://github.com/pod4lib/aggregator/pull/550

[...]

PR #550, opened 2022-04-05
Initial Docker setup for full app

thatbudakguy:
See #532

this doesn't work yet, but hopefully the `docker-compose.yml` is helpful for whoever comes along next. also needs info in README about how to use it.


ktamaral:
@Hurlyburly2 and I made some updates to this branch and added instructions to get this working fully.

ktamaral:
@thatbudakguy I couldn't add you as a reviewer since it was your PR in the first place but please take a look when you have a chance and let us know if you have any comments.

corylown (commented):
Thanks for working on this! The only issue I'm having is that it doesn't look like Sidekiq is able to connect to Redis. When I try to go to http://localhost:3000/sidekiq I see:

```
Redis::CannotConnectError
Error connecting to Redis on 127.0.0.1:6379 (Errno::ECONNREFUSED)
```

README.md
```diff
@@ -80,3 +80,12 @@ bundle exec rails server  # or other commands
...
 ## Deployment
 Deployment is setup using Capistrano using standard [Stanford Digital Library Systems and Services practices](https://github.com/sul-dlss/DeveloperPlaybook/blob/master/best-practices/deployment.md#ruby-applications).
+
+## Getting Stated with Docker
+
+1. `docker-compose -f docker-compose.yml up`
+2. Exec into the running worker container: `docker exec -it aggworker bash`
+3. Inside the worker container, run: `./bin/setup`
+4. Exec into the running app conatiner: `docker exec -it aggapp bash`
+5. Inside the app container, run: `./bin/rails db:migrate`
```
thatbudakguy, line 90: tested this and i think it's sufficient to just do `docker compose exec aggapp ./bin/setup` instead of steps 2-5. you also don't need `container_name` set in the compose file to do this (although i can see why it might be helpful)

thatbudakguy (commented):
tested this out locally; i think we still need a solution for steps 2 and 3 of ["emulating production" in the README](https://github.com/pod4lib/aggregator#emulating-production) (i.e. getting the app to talk to redis/sidekiq). most likely we need to make `config/cable.yml` and `config/environments/development.rb` look to see if `REDIS_URL` is defined and then take action based on that. let's hold off merging this until we can get that working; i will work with @corylown and whoever else to figure it out.

README.md
Hurlyburly2: Katie found that she had to run the db:migrate as well, but that may have been a result of some in development work. With a fresh clone and fresh containers, I also found that `./bin/setup` was sufficient. We'll keep an eye on it just in case, but I think you're right in that it can be kept out.

corylown (approved):

Issue #519, opened 2022-03-30
Customize org section on homepage


ggeisler:
The new homepage design in #470 suggests adding a full-width section just above the footer that will either:

- Offer a Login button and a contact link, when the user is not currently logged-in
- Display a summary of the user's organization status with some quick links to tasks/pages, when the user is logged-in

## Public user (user not logged-in)

<img width="1196" alt="Screen Shot 2022-03-30 at 9 51 43 AM" src="https://user-images.githubusercontent.com/101482/160889798-6207dc3c-0304-436d-965b-6dc19746f1c3.png">

- [x] All elements in this section should be centered in the section
- [x] `Already a POD user?` as an `<h2>` with a `.h3` class (at least to start)
- [x] Login button - takes the user to the same page as the "Login" link in the topbar
- [x] `Interested in using POD data?` as an `<h2>` with a `.h3` class (at least to start)
- [x] `Check out the POD wiki or contact someone@institution.org` where the "POD wiki" link is to https://github.com/pod4lib/aggregator/wiki and the email link is TBD (perhaps @JohnMarkOckerbloom can suggest what the main contact email address for the project should be, or decide if we don't want to offer that here)

## Org member (Provider org)
When a user is logged-in, we know the organization they're associated with so can present some customized info in this section:

<img width="1196" alt="Screen Shot 2022-03-30 at 9 48 07 AM" src="https://user-images.githubusercontent.com/101482/160896796-780c674d-d32b-457a-8eb3-f43a236bcdee.png">

**A:**
- [x] Organization name (and logo image, if it exists) as an `<h2>` with a `.h3` class (at least to start)
- [x] `Most recent upload was on [date and time]` - using last upload info for that organization

A new row below the org header/text. Three columns; we might have to do some trial and error to see how they lay out with real data to decide how much left and right margin to give the row and when the columns should take full width at smaller breakpoints. I can help with that if necessary when the basic code and real data is in place.

**B:**
- [x] `Latest uploads` as an `<h3>` with a `.h4` class (at least to start)
- [x] List of the the filenames of the last three files uploaded by the organization, along with their file status icon, following the pattern we're using in #508. Filenames should not be linked here, I don't think.
- [ ] `View all stream files ...` as a link to the provider page, default stream, Files tab.

**C:**
- [x] `Processing status` as an `<h3>` with a `.h4` class (at least to start)
[...]

Issue #470, opened 2022-03-11
Revise homepage


ggeisler:
The current homepage is out-of-date in several ways (MVP references, deprecated application name) and omits info that could be helpful to new and returning users. A complete redesign of the homepage would be worthwhile and likely not a huge amount of work.

ggeisler:
Below is a proposed redesign of the homepage. In this ticket I'll focus on the new structure of the homepage and the elements that are common to all users. I'll create a separate ticket to focus on the details of what I've labelled section D below, since that seems like a bunch of detailed work that could be done after getting the basic restructuring done (don't want to make this ticket too complicated).

<img width="1196" alt="Screen Shot 2022-03-29 at 2 59 58 PM" src="https://user-images.githubusercontent.com/101482/160715019-15fbd282-b3ea-4505-a14f-30f74106b92f.png">

Above is an example for an organization member who is logged in.

**A**: Re-label the topbar main navigation elements (see additional mockups below for examples not shown above):
  - The current "Organizations" link becomes "Providers" and points to the same page
  - Add a new link "Data" that points to a new page focusing on tasks a data consumer cares most about
  - The current "Activity" link is only shown to users in "Superuser" mode
  - A new link "Users" is only shown to users in "Superuser" mode (goes to the same page the "Manage users" link the user menu goes to)
  - A public/not-logged-in user should not see any of these navigation links
 
**B:** Update the hero title and text.
  - `POD Aggregator`
  - `The POD Aggregator is an open source system to aggregate MARC bibliographic and holdings data for institutions that participate in the Platform for Open Data (POD) initiative.`

Replace the current content between the hero and footer sections with the content labelled "C" and "D" in the mockup above.

**C:** Two equal-width sections that take 50% of the page width. On mobile, they should each take 100% width.
- Use the same border and background colors as used in the site footer
- The headings should be `h2`s but let's try formatting them with the class `.h3` so they're a bit smaller
- If the user is logged in, it would be nice if the entire section of each (not just the heading but the entire block) links to the same place as the "Providers" and "Data" links in the topbar navigation. But if the user is not logged in, those links shouldn't be active.

- `POD Providers`
- `There are currently x data providers. The most recent uploads were made by y1, y2, and y3 on z.` where:
  - `x` = number of data providers
  - `y` = the names of providers who made uploads on the most recent day uploads were made
[...]

Issue #531, opened 2022-03-31
Set up tab navigation for main provider-related pages


ggeisler:
Related to #527.

<img width="1062" alt="provider-page-updates 4" src="https://user-images.githubusercontent.com/101482/161166816-4c911987-ae6b-44dc-badf-a55ce02025dc.png">

---

The provider page currently contains several distinct chunks of stream-related information that would benefit from being more clearly distinguished from each other, so the user can focus on the chunk of information most relevant to their current task.

The goal is to create a tabbed navigation to enable the user to see these distinct chunks of stream-related information and easily move between them. 

- [x] The tabs container should be horizontally-centered on the page, below the new container with the stream file and record counts.
- [x] We can experiment with styling, though I'd suggest trying `.nav-pills` to start. 
- [x] The order and label of the tabs is: `Files`, `Normalized data`, `Processing status`, and `Profile`.
- [x] "Files" is the default tab/page when the user selects links elsewhere on the site for a specific Provider's page.
- [x] The "Processing status" tab should not be displayed to a user who is an org member or org admin and is looking at the provider page of an organization not their own. (A user acting as a superuser should see all tabs regardless of their organization affiliation.) Example: I'm a Stanford user looking at Duke's provider page:

<img width="1062" alt="Screen Shot 2022-03-31 at 4 24 48 PM" src="https://user-images.githubusercontent.com/101482/161166930-4fda155e-e050-47f7-8e0c-962baf52180b.png">

---

## Examples
I'm not sure if we need a separate ticket for this, but for reference these are the chunks of information that go with each tabbed page:

### Files
The current stream files table, but without the totals row:

<img width="1062" alt="Screen Shot 2022-03-31 at 4 18 34 PM" src="https://user-images.githubusercontent.com/101482/161166855-725ef853-0fec-4f18-b98b-f598771d6e72.png">

---

### Normalized data
The normalized data section of the current provider page (we might reformat the table differently, but that would be a separate issue):

<img width="1062" alt="Screen Shot 2022-03-31 at 4 18 44 PM" src="https://user-images.githubusercontent.com/101482/161166872-117b47ba-a889-4848-b4e3-a155d942a844.png">

---

### Processing status
Just the processing status info. There is an [existing ticket about displaying this in a new way](#317).

[...]

PR #555, opened 2022-04-07
Update the home page

thatbudakguy:
looking for advice on further steps to take re: the structuring of `PagesController#home`. rubocop doesn't like it and maybe there is a better place for the logic.

see #519 and #470 for context.

### DIFF
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -5,7 +5,24 @@ class PagesController < ApplicationController
   skip_before_action :authenticate_user!
   skip_authorization_check
 
-  def home; end
+  def home
+    # total providers and most recent upload
+    @providers = Organization.where(provider: true).count
+    @last_upload = Upload.order(created_at: :asc).last
+
+    # total number of MARC records & unique records (by 001 field)
+    @total_records = MarcRecord.count
+    @unique_records = MarcRecord.select('DISTINCT marc001').count
+
+    return if current_user.blank?
+
+    # user's organization and most recent 3 files in its default stream
+    @organization = current_user.organization
+    @last_org_files = @organization.default_stream.files
+                                   .limit(3)
+                                   .reverse_order
+                                   .flat_map(&:attachments)
+  end
 
   def api; end
 
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -6,22 +6,7 @@ class PagesController < ApplicationController
   skip_authorization_check
 
   def home
-    # total providers and most recent upload
-    @providers = Organization.where(provider: true).count
-    @last_upload = Upload.order(created_at: :asc).last
-
-    # total number of MARC records & unique records (by 001 field)
-    @total_records = MarcRecord.count
-    @unique_records = MarcRecord.select('DISTINCT marc001').count
-
-    return if current_user.blank?
-
-    # user's organization and most recent 3 files in its default stream
-    @organization = current_user.organization
-    @last_org_files = @organization.default_stream.files
-                                   .limit(3)
-                                   .reverse_order
-                                   .flat_map(&:attachments)
+    @overview = Overview.new(current_user)
   end
 
   def api; end
--- a/app/models/overview.rb
+++ b/app/models/overview.rb
@@ -15,7 +15,8 @@ class Overview
 
   # most recent upload to default stream by any provider organization
   def last_upload
-    Upload.includes(:organization, :stream)
+    @last_upload ||=
+      Upload.includes(:organization, :stream)
           .where(organization: { provider: true, public: true })
           .where(stream: { default: true })
           .order(created_at: :desc)
--- a/app/models/overview.rb
+++ b/app/models/overview.rb
@@ -56,6 +56,6 @@ class Overview
 
   # public organizations that provide data to the aggregator
   def providers
-    Organization.where(provider: true, public: true)
+    @providers ||= Organization.where(provider: true, public: true)
   end
 end
--- a/app/models/overview.rb
+++ b/app/models/overview.rb
@@ -17,10 +17,10 @@ class Overview
   def last_upload
     @last_upload ||=
       Upload.includes(:organization, :stream)
-          .where(organization: { provider: true, public: true })
-          .where(stream: { default: true })
-          .order(created_at: :desc)
-          .first
+            .where(organization: { provider: true, public: true })
+            .where(stream: { default: true })
+            .order(created_at: :desc)
+            .first
   end
 
   # total number of MARC records in provider default streams
--- a/app/models/overview.rb
+++ b/app/models/overview.rb
@@ -1,0 +1,60 @@
+# frozen_string_literal: true
+
+# At-a-glance information for the POD aggregator and a single organization
+class Overview
+  attr_reader :organization
+
+  def initialize(user)
+    @organization = user.organization if user.present?
+  end
+
+  # total number of public organizations that provide data
+  def provider_count
+    providers.count
+  end
+
+  # most recent upload to default stream by any provider organization
+  def last_upload
+    Upload.includes(:organization, :stream)
+          .where(organization: { provider: true, public: true })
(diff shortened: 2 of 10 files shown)


config/locales/en.yml
```diff
@@ -61,6 +61,48 @@ en:
...
+        heading: POD Data
+        body_html:
+          zero: 'POD Aggregator currently holds no records.'
+          one: 'POD Aggregator currently holds <b>1 record</b>.'
+          other: 'POD Aggregator currently holds a total of <b>%{total_records} records</b>, of which <b>%{unique_records} are unique</b>.'
+      login:
+        heading: Already a POD user?
+        button: Login
+        contact_heading: Interested in using POD data?
+        contact_html: Check out the <a href='%{wiki_url}'>POD wiki</a>.
```
anarchivist, line 83:
```suggestion
        contact_html: Check out the <a href='%{wiki_url}'>POD wiki</a> or contact <a href='mailto:pod-support@lists.stanford.edu'>pod-support@lists.stanford.edu</a>.
```

It's fine by me if we want to move that email into an i18n key or some other config option.

app/views/pages/home.html.erb
```diff
@@ -1,71 +1,136 @@
...
+        <ul class="list-unstyled">
+          <% if @organization.provider? %>
+          <li class="my-3"><%= link_to t('.organization.actions.upload_file'), new_organization_upload_path(@organization) %></li>
+          <li class="my-3"><%= link_to t('.organization.actions.view_default_stream'), organization_stream_path(@organization, @organization.default_stream) %></li>
+          <% else %>
+          <li class="my-3"><%= link_to t('.organization.actions.harvest_data'), organization_path(@organization) %></li>
+          <li class="my-3"><%= link_to t('.organization.actions.data_docs'), 'https://github.com/pod4lib/aggregator/wiki/Harvesting-using-ResourceSync' %></li>
+          <% end %>
+          <% if can? :manage, @organization %>
+          <li class="my-3"><%= link_to t('.organization.actions.edit_org'), edit_organization_path(@organization) %></li>
```
marlo-longley, line 113: @thatbudakguy I pulled the branch locally and was clicking around, noticed with the now-merged changes to the Manage Org page, the `edit_organization_path` no longer exists, the good link would be: `organization_users_path(@organization)` . I wonder if we should add tests to make sure the homepage links are all valid? Or maybe this is just a one-time thing since  a lot of simultaneous changes are happening.

thatbudakguy: oh yeah when i was writing that it didn't exist, good catch. i can update these now!

app/controllers/pages_controller.rb
```diff
@@ -5,7 +5,24 @@ class PagesController < ApplicationController
...
+  def home
+    # total providers and most recent upload
+    @providers = Organization.where(provider: true).count
+    @last_upload = Upload.order(created_at: :asc).last
+
+    # total number of MARC records & unique records (by 001 field)
+    @total_records = MarcRecord.count
+    @unique_records = MarcRecord.select('DISTINCT marc001').count
+
+    return if current_user.blank?
```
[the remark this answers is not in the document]
corylown: I still think it might be nice to create a model/class to contain all the logic needed to fetch data to build this home page view. You could pass it the current user and whatever else it needs and then you'd be able to pass one instance of this new class to view with methods for fetching the data to you need.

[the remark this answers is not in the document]
cbeer: >  create a model/class to contain all the logic needed to fetch data 

👍 . I think w'll find some of the data fetching will be pretty expensive, so this will give us an easy place to cache it too.

app/views/pages/home.html.erb
thatbudakguy: fyi I updated this one, but i think one or two of the others link to tabs on the organization page, and are thus blocked on #531

app/controllers/pages_controller.rb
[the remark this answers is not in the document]
thatbudakguy: extracted this stuff to an `Overview` model and made that rely on pre-calculated statistics based on @corylown's advice. looking for more 👀 to see if anything else needs perf tuning; currently I can't ssh to prod to run ActiveRecord queries myself for some reason

app/views/pages/home.html.erb
```diff
@@ -1,71 +1,134 @@
...
-          </ul>
-        </div>
       <% end %>
+      <div class="my-2 my-md-0">
+        <h3>
+          <%= t('.organization.actions.heading') %>
+        </h3>
+        <ul class="list-unstyled">
+          <% if @overview.organization.provider? %>
+          <li class="my-3"><%= link_to t('.organization.actions.upload_file'), new_organization_upload_path(@overview.organization) %></li>
```
corylown, line 104: I think we need a permission check here because org members cannot currently upload files. I'm not 100% sure what behavior we want, but currently they cannot upload files.

app/views/pages/home.html.erb
```diff
@@ -1,71 +1,134 @@
...
+      <% if @overview.organization.provider? %>
+      <div class="my-2 my-md-0">
+        <h3>
+          <%= t('.organization.jobs.heading') %>
+        </h3>
+        <ul class="list-unstyled">
+          <% if @overview.active_org_jobs.present? %>
+          <% @overview.active_org_jobs.each do |status, jobs| %>
+          <li class="my-3">
+            <%= bootstrap_icon(Settings.job_status_group[status].icon_class, class: "pod-icon pod-job-tracker-status #{status} me-2")%>
```
corylown, line 84: Might be nice if the needs attention icon displayed in yellow like it does in other contexts.

app/models/overview.rb
```diff
@@ -0,0 +1,60 @@
...
+  end
+
+  # total number of public organizations that provide data
+  def provider_count
+    providers.count
+  end
+
+  # most recent upload to default stream by any provider organization
+  def last_upload
+    Upload.includes(:organization, :stream)
```
corylown, line 18: I'd be tempted to memoize these so the database calls only have to happen once per request.
```suggestion
    @last_upload ||=
      Upload.includes(:organization, :stream)
```

app/models/overview.rb
```diff
@@ -0,0 +1,60 @@
...
+    @organization.default_stream.job_tracker_status_groups.filter do |_status, jobs|
+      jobs.count.positive?
+    end
+  end
+
+  private
+
+  # public organizations that provide data to the aggregator
+  def providers
+    Organization.where(provider: true, public: true)
```
corylown, line 58: I'd memoize this too.
```suggestion
    @providers ||= Organization.where(provider: true, public: true)
```

corylown (commented):
This is a huge improvement over the current home page. Looks really great. A few inline suggestions.

app/models/overview.rb
```diff
@@ -0,0 +1,61 @@
+# frozen_string_literal: true
+
+# At-a-glance information for the POD aggregator and a single organization
+class Overview
```
corylown, line 4: Any thoughts on this class name? `Overview` seems a bit generic but we already use Dashboard for something else in the app. Kind of seems like a `UserDashboard` to me. Maybe we can deal with this later. Just thinking...

app/views/pages/home.html.erb
thatbudakguy: What's the right way to do that? Something like `if can? :create, Upload, { organization: @overview.organization, stream: @overview.organization.default_stream }`?

corylown: I think this will work: `can? :create, @overview.organization.default_stream.uploads.build`

app/views/pages/home.html.erb
thatbudakguy: Good catch; I think the easiest way to fix that was just adding another option to the CSS. Should be good now

app/views/pages/home.html.erb
thatbudakguy: I think that did it, thanks!

corylown (approved):

Issue #544, opened 2022-04-04
Add "contact" field to New organization form


ggeisler:
The current new organization form at `/organizations/new` does not currently include a contact address field. However, through #515 and related discussions we decided to change "organization email contacts" to be an optional,  single-valued field. 

As a single-valued field representing a basic attribute of organization, I think it make sense to move this field from the current organization contacts location to a field editable through the "new organization" and "organization details" forms. This ticket focuses just on the new organization form accessed at `/organizations/new`

<img width="981" alt="Screen Shot 2022-04-04 at 3 44 47 PM" src="https://user-images.githubusercontent.com/101482/161645264-3039b469-6b0b-4143-bfba-1b91087ab495.png">

---

- [ ] Add a new "contact" attribute to organization and include it in the new organization form, just above the icon input, as shown above. Details:

Field label: `POD contact for this organization`
Hint: `An email address other POD users can use to contact this organization.`

**EDIT**: Note that the hint text above doesn't match the mockup exactly but that is the hint text we should use, rather than what you see in the mockup (update reflects the decision to only use email address for a contact.)

corylown:
Do we really need to support a url pointing to a form or contact info or can we just use an email address?

cbaksik:
We would like an email field. A separate URL field might be nice - to direct users the "right" part of the library website. 

We plan to use a single email that goes to a support team, and not to an individual person.

JohnMarkOckerbloom:
I've sent out email to the pod-tech list asking whether anyone needs to use a contact URL for this, or whether a single email address is sufficient.  I'll see if I get any replies between now and Thursday's meeting, but at least for now I have not, and at least for Penn (per Bob) and for Harvard (per Corinna's comment above) a single email address would be fine.  I'll wait until after that meeting before making a definite decision, though.

JohnMarkOckerbloom:
After a suitable interval, I have not heard from anyone needing a URL contact field.  So I think a single email address should be fine for now.

JohnMarkOckerbloom:
I've answered the question above, but the implementation of the form addition needs to be done by someone else, so this issue should probably be reassigned to that person.

corylown:
@JohnMarkOckerbloom I've unassigned you and queued this up for sprint 4

corylown:
Relates to #545

Issue #515, opened 2022-03-28
Should organization contact email address be a single-valued, required field?


ggeisler:
Currently we allow an organization to add more than one contact email address but it is optional.

I realize we're still ramping up to realistic use of POD but as we move towards real production in this workcycle I wonder if we can look at this field again. Specifically, it seems to me that it is not unreasonable to:

- **Make contact email address a required field**. Since POD users will be able to look at other providers files, etc. it seem reasonable that a POD user might have a question for a given provider about their data. They should have a clear way to ask their question of the appropriate person/mailing list at the organization.

- **Change contact email address to a single-valued field**. Providing multiple email addresses will pose a challenge to putting the contact email address in a prominent location in the upcoming redesign, but just as importantly, requiring the organization to choose a single email address seems like a good way forcing the organization to think about how they want to handle POD-related requests (i.e., creating or using an existing mailing list rather than just adding a bunch of people and making the POD user with a question try to decide which one to email or to email everyone, etc.).

For reference, I believe that currently only Stanford and Yale have provided a contact email address. Stanford is pointing to a single mailing list, while Yale has listed two individual email accounts.

And here is the proposed page header where the contact email address will be more prominently displayed:

<img width="1197" alt="Screen Shot 2022-03-28 at 1 46 24 PM" src="https://user-images.githubusercontent.com/101482/160488268-985b6cc6-f68f-49f0-9434-92f974a0d554.png">

ggeisler:
@JohnMarkOckerbloom Thanks for the comments. The actionable steps I take from this are:

- You will verify that organizations are willing to provide one, and only one, contact email or link (I don't know how we should balance "asking" versus "informing"; I think this is something we **should do** unless someone has a really good reason not to, rather just a casual "we'd rather provide multiple emails")
- I'll ticket an update for the UI that displays the contact info so it is displayed as a link (to either an email address or a URL)
- I'll include relevant updates to the form used for managing the contact info (you don't currently see it from the Edit org page because it is in a section at the very bottom of the main org page; but that's been updated in the new redesign plan)

As you said, non-logged in users only see the homepage and we don't plan to display the contact address for an organization there, so there's no issue with non-POD users seeing this.

JohnMarkOckerbloom:
Thanks!  I'm thinking the best way to touch base with our providers is to send an email to the pod-tech list (since I suspect more people read that that are on the Slack instance or follow Github issues).  I'll plan to draft and send out something later today or tomorrow.

[...]

PR #577, opened 2022-04-11
Support routing, permisisons, and UI for new Manage Organization functionality

### LINKED ISSUE #528: Update layout and contents of Provider page banner
Related to #527.

provider-page-updates

---

**A: Organization code** 
- [ ] Update the current label to `MARC organization code` from the current `Code`
- [ ] Reposition the code label and value so that it follows the organization name (maybe `2rem` horizontal space between them)
- [ ] If the organization is consumer-only, omit display of the MARC code, since the consumer-only org won't have one

The organization contact email value is currently embedded in a card structure but this suggestion is to just use the **value** in the new location. We'll display the edit functions for this element in a revised organization details page.

**B: Organization contact info**
- [ ] Add a label `Contact:` below and left-aligned with the MARC code label
- [ ] Move the value of the organization contact email(s) from its current location to follow the Contact label

This design assumes we're constraining the organization contact email to one value, however, we don't currently do this. There's a [ticket to resolve this](#515). In the meantime, we currently have one organization that does have two contact email addresses. If that isn't resolved before someone works on this I suggest we just display them one after the other, comma-separated (or possibly break to new line for each; I'm just trying to keep the org banner section from being too tall).

**C: Organization details link**
- [ ] Change the current "Edit" button to a link. The label and its visibility depend on the user's role.
- [ ] Superuser or org admin user: link label is `Manage organization`
- [ ] Org member (of current provider): link label is `View organization details`
- [ ] Org member (of different provider): link is not displayed

See #534 for details of how we want to redo the Manage/view organization page.

### LINKED ISSUE #534: Set up new structure for Manage/View Organization (Provider) page(s)
Currently, to view or edit some details about a given organization (users, contact emails) you have to scroll to the bottom of the organization page, below the list of files and normalized data. To view or edit other details (org name, MARC code, MARC field details), you use a different path, the Edit button in the organization banner.

The redesign of the provider page (#527) proposes moving all of this organization-specific, non-stream related information to a single location, accessed by selecting either a "Mange organization" (for org admins) or "View organization details" (org members) link, as described in section C of #528. This is the "Manage organization" version of the page org **admins** should see:

Screen Shot 2022-04-04 at 12 20 10 PM

---

Org **members** see something very similar, except the page heading is "View organization" and the main content of the tabbed pages will be read-only:

Screen Shot 2022-04-04 at 12 21 59 PM

---

Note that we want to use the same structure for organizations that are **consumer-only**, but things will look a little different. I'll create a separate ticket about that.

The tasks below assume we're using the current organization edit page (`organizations/[org]/edit`) as the basis for the updated page, although I'm not suggesting that that is the way we should actually implement it.

- [x] **A:** Replace the current "Edit" button (only currently seen if org admin, I think) with a link labelled "Default stream." The link leads back to the provider page, default stream active, Files tab active. Both org members and org admins should see this link.
- [x] **B:** Replace the "Editing organization" header with either "Manage organization" (when user is org admin/superuser) or "View organization" (when user is org member).
- [x] **C:** Tabs structure should visually match what we do on the provider page (see #531): horizontally-centered, etc.
- [x] "Users" is the default tab when the user lands on this page. Move the "Users" section currently at the bottom of the provider page to be the main content below the tabs when the "Users" tab is active.
- [x] Move the "Access tokens" section currently at the bottom of the provider page to be the main content below the tabs when the "Access tokens" tab is active.
- [x] Move the first part of the form on the "Editing organization" page to be the main content below the tabs when the "Organization details" tab is active. This will also require making it a standalone form (currently it is grouped with the MARC details) and displaying a read-only version for org members (only org admins/superusers can edit the form). I'll create a separate ticket about that.
- [x] Move the second part of the form on the "Editing organization" page to be the main content below the tabs when the "Provider details" tab is active. This will also require displaying a read-only version for org members (only org admins/superusers can edit the form). I'll create a separate ticket about that.

### LINKED ISSUE #539: Update the Manage/View organization users tab/page
Related to #534.

There are two parts to this:

- Change the current structure of the Organization Users list to a table (so we can display the different user attributes better)
- Ensure that when this new table is viewed by org members (rather than org admins/superusers), the edit components are not available

## Change the current structure of the Users list to a table
Organization users are currently displayed in a list, within a Bootstrap card component. Here's how I suggest they should be displayed (this is what the **org admin/superuser** sees):

Screen Shot 2022-04-04 at 2 14 29 PM

---

- [x] Remove the Bootstrap card component the user-related elements are currently part of. That made sense when we were showing a lot of different things on the same page, but now that we're focusing each page on a single thing, there's no need for the extra borders and header of the card component.
- [x] **A:** Use the existing "Invite" button but position it directly below the page heading "Manage users", left-aligned in the content area
- [x] **A:** Re-label the Invite button `Invite new user to organization`
- [x] **B:** To display the user data, switch to a table layout from the current unordered list
- [x] **B:** Break out each user field to its own column. The mockups in this ticket reflect our current state where every user has an email address but name and title are optional.
- [x] **B:** Add a new column "Role" with the user's role (`Member` or `Admin`). (We don't currently show this in the org user list, but I think it could be handy for org users to see whose in their org is an admin, in case they need some admin-only actions to be taken.)
- [x] **C:** Create a final column labeled `Actions`. Move the Remove buttons to this column, centered in the column, and change the label to `Remove user`.

## Ensure page is read-only for non-admins
If the user is an org member, the page should look more like this:

Screen Shot 2022-04-04 at 2 16 01 PM

---

Where we:

- [x] Don't display the Invite button
- [x] Don't display the Actions column and its Remove user buttons

I believe all of the above is the same whether the organization is a provider or consumer-only; no special case for consumer-only on this page.

### LINKED ISSUE #541: Update the Manage/View organization Access tokens tab/page
Related to #534.

Screen Shot 2022-04-04 at 2 42 25 PM

- [x] Remove the Access tokens card currently in the Organization management section at the bottom of the organization page. All this currently does is provide the user with a link to a dedicated Manage tokens page, which in the redesign will be available from the Access tokens tab on the Manage/View organization page.
- [x] On the Access tokens tab of the Manage/View organization page, display the current Manage tokens page content, as shown above.
- [x] (Optional, but would be helpful to the user if we have time to do it) Update the current Manage tokens page content mentioned above by adding a new primary button with the label "Copy". This button copies the associated access token string to the user's clipboard, so they don't have to manually copy it from the input.
- [x] If the user is a **org member**, and not an org admin, remove the "Create token" button they currently see (based on our new roles and permissions spreadsheet, only org admins should be able to create new access tokens).

The mockup at the top of this ticket is what the org admin sees. Here's the org member view of the page:

Screen Shot 2022-04-04 at 2 38 58 PM

---

### Bonus points:

- [x] Add the `disabled` HTML option to the input with the token value (since the value isn't editable, this will make it easier/more obvious for the user if they copy directly from the input -- unless I'm missing something and the token value should be editable).
- [ ] Add the Font Awesome `copy` icon to the button. We aren't generally following a style of having icons on buttons in the application, but this seems like a special case where it would be useful to give the user an additional hint that the button can be used to easily copy the token value.

The two additions above would make things look more like:

Screen Shot 2022-04-07 at 4 58 04 PM

### LINKED ISSUE #542: Update the Manage/View organization Provider details tab/page
Related to https://github.com/pod4lib/aggregator/issues/534.

## Org admin version of the page
The last task in #534 suggests moving the second part of the form on the "Editing organization" page to be the main content below the tabs when the "Provider details" tab is active, as shown below:

Screen Shot 2022-04-04 at 2 43 22 PM

Because we're splitting the current Editing organization form into two, on this page and the Organization details page, we can customize the label of the submit button on each page. Here:

- [x] Label the submit button `Update provider details`

I think that's all that needs to be done for the **org admin** view of the Provider details page.

## Org member version of the page
We don't currently have a read-only view of this form for org members, but I think it would be useful for at least consistency purposes if the org member can see the values of this form, even if they can't update the form.

So, when the user is an **org member**, instead of displaying the editable form, we could show the current values in a description list:

Screen Shot 2022-04-04 at 2 37 19 PM

---

- [x] Omit the submit button in the org member's version of this page.

- [x] Put provider details form fields into a definition list. Use a `dl` element and put the value of each field label in a `dt` and the field value in a `dd`.

- [x] Indicate when a field has no current value. If for some reason there isn't a current value for any field, I suggest we display the string `no value provided` in italics so it's clear what the situation is:

Screen Shot 2022-04-04 at 3 11 40 PM

### LINKED ISSUE #546: Update the Manage/View organization Organization details tab/page
Related to https://github.com/pod4lib/aggregator/issues/534.

## Org admin version of the page

The next to last task in #534 suggest moving the first part of the form on the "Editing organization" page to be the main content below the tabs when the "Organization details" tab is active:

Screen Shot 2022-04-04 at 4 27 59 PM

---

Because we're splitting the current Editing organization form into two, on this page and the Provider details page, we can customize the label of the submit button on each page. Here:

- [x] Label the submit button `Update organization details`

We're adding "organization contact" as an optional, single-valued field, as described in #544 and #545. It therefore should be added to the form on this page, so an org admin can update their contact value.

- [ ] Use the updated (through #544) organization form on this page (or if it makes sense to maintain an independent form here, add the "POD contact for this organization" field as described in #544), so it ends up looking like the example above, with the red arrow (except the strings are Field label: `POD contact for this organization` and 
Hint: `How should other POD users contact this organization? Provide an email address or URL.`).

## Org member version of the page
We don't currently have a read-only view of this form for org members, but I think it would be useful for at least consistency purposes if the org member can see the values of this form, even if they can't update the form.

So, when the user is an **org member**, instead of displaying the editable form, we could show the current values in a description list:

Screen Shot 2022-04-04 at 4 38 24 PM

---

- [x] Omit the submit button in the org member's version of this page.

- [x] Put organization details form fields into a definition list. Use a `dl` element and put the value of each field label in a `dt` and the field value in a `dd`.

- [x] If for some reason it's troublesome to include the icon here, I don't think it's essential, since if it is defined the user should be able to see it in the top banner of the page.

- [ ] Indicate when a field has no current value. If for some reason there isn't a current value for any field, I suggest we display the string `no value provided` in italics so it's clear what the situation is:

Screen Shot 2022-04-04 at 4 31 22 PM

marlo-longley:
Fixes #534 and its sub tasks:

fixes #528
fixes #539 
fixes #541 
fixes #542
fixes #546 

Setting this as a draft PR for now. I realize this is pretty huge with a lot of file changes, and only 1 commit! I didn't feel good about splitting up the changes though - many of the view changes rely on adding new links to new routes, which rely on the controllers, etc etc -- everything was very tightly wrapped up! But I am very open to suggestions on how to make this more manageable. 

I'm also interested in getting review on my testing - want to be thorough with this many changes. One thing I had no idea how to test was the custom javascript file I created. Not sure if we need to do that.

I wanted to get the draft reviewed and in good shape before dealing with the merge conflicts. I'm open to suggestions on the best way to tackle that, too.


marlo-longley:
Some screenshots of the views when logged in as org owner
Screen Shot 2022-04-11 at 19 38 15
Screen Shot 2022-04-11 at 19 38 21
Screen Shot 2022-04-11 at 19 38 29
Screen Shot 2022-04-11 at 19 38 38

app/assets/stylesheets/organizations.scss
```diff
@@ -16,6 +16,9 @@
   }
 }
 
+// Access bootstrap color utility variables
+@import 'bootstrap';
```
corylown, line 20: Is this needed? We're already importing bootstrap in application.scss.

corylown:
@marlo-longley I think this looks good to me overall, but I have a few questions:

1. There's still an organization management section on the default stream page that lists users. Should we remove this as part of this PR or do you want to handle it separately?
2. I'm assuming you're intention is to handle moving the contact info from the default streams page to the organization details tab as part of a separate issue (and having it show up in the organization header)? Just wanted to check.
3. The tabs are a bit fiddly to use. I seem to have to click on the text to get it to work. Clicking on the tab outside the text doesn't seem to work. Is this a bootstrap 5 thing? Is there a way to fix it so the whole tab is clickable?
4. This may be related to 3. When I hover over an unselected tab but not over its text the text and the tab is blue and so the text is unreadable. Wondering if we should fix this.

marlo-longley:
Hey @corylown thanks for those comments. 

1. Yeah, I figured this PR was big enough, and the next thing I plan to work on is to overhaul the Providers page, which should take care of anything redundant.

3. Thanks for pointing that out. I thought the contact info wasn't implemented yet, so I left that out of the header. There is an existing issue for adding it to the form https://github.com/pod4lib/aggregator/issues/544 - seems related / blocked by https://github.com/pod4lib/aggregator/issues/515 ? I may be missing some context.

3 + 4: Let me improve the tab experience and make the whole tab, not just the inner link, clickable.

corylown:
@marlo-longley I think if you address 3 & 4 this is merge-able and we can deal with 1 & 2 as separate issues.

app/assets/stylesheets/organizations.scss
marlo-longley: I had undefined variables unless I did this - not sure about the deeper implications for asset loading.

marlo-longley:
@corylown I addressed 3+4 - the tabs are much better! And now can be a solid basis for other tabs in the app. Thanks.

app/javascript/copy_to_clipboard.js
```diff
@@ -0,0 +1,17 @@
+window.addEventListener('load', function() {
```
corylown, line 1: @marlo-longley this doesn't seem to work if there's more than one token on the page. I can copy one of them, but if I then try to copy the other one it doesn't work.

marlo-longley: Ohhh that makes total sense. I obviously should have but only had 1 token on my local and didn't account for that. Thank you! @corylown

Issue #560, opened 2022-04-07
Provider page > Processing status


cbaksik:
In the Processing status section #531 move "Needs attention" above "Currently active" and have "Needs attention" open by default.

ggeisler:
👍 This makes sense to me. 

Probably obvious, but "Needs attention" tab should only be open by default if it includes one or more jobs. Otherwise, "Currently active" should be open by default, if it includes one or more jobs. Otherwise, "Completed" should be open by default, if it includes one or more jobs. If no tab includes jobs, no tab should be open.

ggeisler:
A new, additional suggestion for the processing status section is in #576. It's completely fine to handle that separate from this, just want to point it out in case the developer working on this ticket wants to tackle that as well, since it involves the same section of code.

ktamaral:
Pull request: https://github.com/pod4lib/aggregator/pull/582

PR #582, opened 2022-04-12
Background job status display tab collapse

ktamaral:
This resolves [560](https://github.com/pod4lib/aggregator/issues/560).

The "needs attention" tab will always display at the top of the list of job groups. If the needs attention tab has any jobs, it will be open by default. Otherwise, the "active" job list will be open by default if it has any jobs. Otherwise the "completed" tab will be open by default if it has jobs. No tabs will be open if there are no jobs.

NOTE: I was able to test this locally to a certain extent but wasn't able to find any organizations that have jobs in the 'needs attention' status. This should work as expected but it may be worth considering creating a job in the needs attention status for testing purposes.

### DIFF
--- a/spec/helpers/application_helper_spec.rb
+++ b/spec/helpers/application_helper_spec.rb
@@ -25,5 +25,20 @@ RSpec.describe ApplicationHelper do
       set_tab_class = helper.job_status_tabs_collapse(false, 3)
       expect(set_tab_class).to be('show')
     end
+
+    it 'applies class to collapse tab when tab expanded is true and job count is positive' do
+      set_tab_class = helper.job_status_tabs_collapse(true, 3)
+      expect(set_tab_class).to be('collapsed')
+    end
+
+    it 'applies class to collapse tab when tab expanded is true and job count is zero' do
+      set_tab_class = helper.job_status_tabs_collapse(true, 0)
+      expect(set_tab_class).to be('collapsed')
+    end
+
+    it 'applies class to collapse tab when tab expanded is false and job count is zero' do
+      set_tab_class = helper.job_status_tabs_collapse(false, 0)
+      expect(set_tab_class).to be('collapsed')
+    end
   end
 end


app/views/streams/_job_status.html.erb
```diff
@@ -1,7 +1,10 @@
 <div class="mb-5">
   <h3><%= t('job_tracker.title') %></h3>
   <div class="accordion" id="jobStatusAccordian">
-    <% stream.job_tracker_status_groups.each do |status_group, jobs| %>
+    <% needs_attention = stream.job_tracker_status_groups.select { |sg| sg == :needs_attention } %>
+    <% status_groups_ordered = needs_attention.merge(stream.job_tracker_status_groups) %>
```
corylown, line 5: Maybe set the order of the status groups in the model so you don't have to mess with it in the view?:

https://github.com/pod4lib/aggregator/blob/main/app/models/stream.rb#L30

app/views/streams/_job_status.html.erb
```diff
@@ -1,7 +1,10 @@
...
+    <% tab_expanded = false %>
+    <% status_groups_ordered.each do |status_group, jobs| %>
       <div class="accordion-item">
         <h4 class="accordion-header" id="heading<%= status_group.to_s.classify %>">
           <button class="accordion-button collapsed" type="button"
```
corylown, line 10: I think you'll need to add logic to add/remove the `collapsed` class on this button too.

app/views/streams/_job_status.html.erb
```diff
@@ -13,8 +16,9 @@
             </span>
           </button>
         </h4>
-        <div id="collapse<%= status_group.to_s.classify %>" class="accordion-collapse collapse"
+        <div id="collapse<%= status_group.to_s.classify %>" class="accordion-collapse collapse <%= !tab_expanded && jobs.count > 0 ? 'show' : 'collapsed' %>"
```
corylown, line 19: Since you're going to need to use this logic in more than one place maybe it would make sense to write a helper method? It could handle the logic needed to support only opening one accordian even if more than one has jobs, etc.

corylown (commented):
A few inline suggestions.

app/views/streams/_job_status.html.erb
ktamaral: OH wow that was so much easier! Thank you

corylown (approved):
Looks good to me!

PR #584, opened 2022-04-12
Issue552 relocate flash message on organization pages

Hurlyburly2:
As discussed in standup, we aren't too concerned about testing location of the element on the page, so no new tests currently. Please let me know if I should do otherwise.

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -19,6 +19,7 @@
   <!-- header -->
   <header>
     <%= render 'shared/nav' %>
+    <!-- if an organization header is displayed, flash messages will instead display in _organization_header.html.erb -->
     <% if !defined? @organization %>
       <div class="container">
         <%= render partial: '/shared/flash_msg', layout: nil %>
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -19,9 +19,11 @@
   <!-- header -->
   <header>
     <%= render 'shared/nav' %>
-    <div class="container">
-      <%= render partial: '/shared/flash_msg', layout: nil %>
-    </div>
+    <% if !defined? @organization %>
+      <div class="container">
+        <%= render partial: '/shared/flash_msg', layout: nil %>
+      </div>
+    <% end %>
   </header>
 
   <!-- main content -->
--- a/app/views/shared/_organization_header.html.erb
+++ b/app/views/shared/_organization_header.html.erb
@@ -14,6 +14,10 @@
       <% end %>
     </div>
 
+    <div class="container">
+      <%= render partial: '/shared/flash_msg', layout: nil %>
+    </div>
+
     <%= yield %>
   </div>
 </div>


app/views/layouts/application.html.erb
```diff
@@ -19,9 +19,12 @@
   <!-- header -->
   <header>
     <%= render 'shared/nav' %>
-    <div class="container">
-      <%= render partial: '/shared/flash_msg', layout: nil %>
-    </div>
+    <!-- if an organization header is displayed, flash messages will instead display in _organization_header.html.erb -->
+    <% if !defined? @organization %>
```
corylown, line 23: Checking for the existence of the @organization instance variable may not be a reliable way to check whether the organization header is rendered. I'd suggest refactoring how the organization header is rendered to address this issue. One way to do this would be to render the organization header within the header tag in `layouts/application.html.erb`. Something like:

```
  <header>
    <%= render 'shared/nav' %>
    <%= yield :org_header %>
      <div class="container">
        <%= render partial: '/shared/flash_msg', layout: nil %>
      </div>
```

You would then wrap the the `organization_header.html.erb` partial in a `content_for` block so it will render when the named yield is called:

```
<% content_for :org_header do %>
  <div class="position-relative overflow-hidden mb-5 pb-2 pt-4 bg-light organization border-bottom">
  #etc.
<% end %>
```

With this approach the org header is always rendered before the flash message so you don't have to worry about trying to figure out if the org header is present or not to conditionally display the flash message in different locations. Happy to pair on this if you have questions.

corylown (commented):
See inline suggestion for a different approach.

app/views/layouts/application.html.erb
corylown: Rails docs on `content_for`:

https://guides.rubyonrails.org/layouts_and_rendering.html#using-the-content-for-method

Hurlyburly2: @corylown Thank you for the detailed tips. I've refactored to use content_for and yield. I also removed the extra `<%= yield %> ` from the organization_header, as I think with the new pattern it doesn't actually do anything. Let me know if this is along the lines of your expectation.

app/views/shared/_organization_header.html.erb
```diff
@@ -1,19 +1,19 @@
...
-      <% if controller_name == 'organizations' %>
-        <%= link_to 'Edit', edit_organization_path(@organization), class: 'align-self-center btn btn-primary' if can? :edit, @organization %>
-      <% end %>
+        <% if controller_name == 'organizations' %>
+          <%= link_to 'Edit', edit_organization_path(@organization), class: 'align-self-center btn btn-primary' if can? :edit, @organization %>
+        <% end %>
+      </div>
     </div>
-
-    <%= yield %>
```
corylown, line 17: I think you still need `yield` here so that content passed to the `render 'shared/organization_header'` blocks will appear, such as here:

https://github.com/pod4lib/aggregator/blob/8c1878467bcc5112b51ca05a4655930fd6635f70/app/views/organizations/show.html.erb#L1

Hurlyburly2: @corylown Tested it out and you're right, the yield has been added back in. I'm not sure if this is standard for you, but this may be a good PR for a squash-and-merge so as not to clutter up the project history with a bunch of commits that were ultimately superfluous.

corylown (approved):

Issue #579, opened 2022-04-12
Links at top of screen in Chrome


bobpersing:
The links in the black bar at the top of the POD screen are displaying correctly in Firefox.  In Chrome, though, they're not correctly spaced, and overlap each other.  Resizing the window doesn't make a difference.

![screenshot-pod stanford edu-chrome-2022 04 12-09_04_28](https://user-images.githubusercontent.com/74777933/162969239-89059cb3-f7a9-4411-a8ce-66e7337e7892.png)
![screenshot-pod stanford edu-firefox-2022 04 12-09_00_58](https://user-images.githubusercontent.com/74777933/162969242-7c96e80c-c750-43ca-81a2-7794eec973b3.png)

Screenshots attached.

ktamaral:
Pull Request: https://github.com/pod4lib/aggregator/pull/582

corylown:
@bobpersing thank you for reporting this. The fix is applied to the main branch of the project and will be deployed to production next Monday (at the latest) with dependency updates.

PR #587, opened 2022-04-12
Add a width percentage value to fix nav overlap

ktamaral:
Fixes [#579 ](https://github.com/pod4lib/aggregator/issues/579)
Tested in Chrome and Firefox on Mac and PC (using Cross Browser Testing platform)

### DIFF
--- a/app/views/shared/_nav.html.erb
+++ b/app/views/shared/_nav.html.erb
@@ -1,7 +1,7 @@
 <nav class="navbar navbar-dark navbar-expand-md bg-dark" role="navigation">
   <div class="container">
     <%= link_to root_url, class: 'navbar-brand' do %>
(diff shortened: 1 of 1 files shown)


corylown (approved):

marlo-longley:
Thanks @ktamaral ! Pretty minor, but I noticed that prior to this commit https://github.com/pod4lib/aggregator/commit/e7059bf0494c6864aea8e50d9f6c86e10059b2b4 the `inline_svg_pack_tag` had a width of 80, which solved the issue and didn't overlap the existing container - thoughts?
(80)
Screen Shot 2022-04-12 at 13 59 45
(33%)
Screen Shot 2022-04-12 at 14 00 00

Also as a bonus wonder if we should add some alt text instead of leaving it blank?

marlo-longley:
@ktamaral Oh I see it was merged ! All good.

Issue #517, opened 2022-03-30
Create new Data page and link from topbar nav and homepage


ggeisler:
I believe our goal is to provide a page that is a clear destination for users who are specifically interested in consuming data (rather than our current approach, which requires a data consumer to find what they want through the Organization pages).

Below is a mockup of the Data page, although this is really more to convey the idea than the exact content. I don't know what the use cases for data consumers are so am not certain what all should be part of this page. I think we could create this page now and link it up to the new navigation and homepage, but we will probably want to populate it with the appropriate content over time as we learn better what data consumer use cases are.

<img width="1195" alt="Screen Shot 2022-04-11 at 2 34 03 PM" src="https://user-images.githubusercontent.com/101482/162837386-c379c565-5224-4b4e-b86c-4728e3d17fca.png">

- This Data page is linked for any logged-in user via the "Data" navigation item in the topbar, as described in #470. In addition, in #470 I suggest the "POD Data" section below the hero block also link to the Data page.

- The Data page should not be accessible to anyone not logged in to POD.

I'll put a Question label on this ticket to encourage people to add use cases and ideas for what should go on this page to comments below.

cbaksik:
Page should include text of or link to the data re-use agreement. Currently it's in a [Google doc](https://docs.google.com/document/d/1EZG7z_A13jxPfHbYhT7Ae4Ays10fhecqfoGJhqrcjD8/edit?usp=sharing) but should probably be elsewhere.

ggeisler:
@cbaksik This sounds like a good idea to me. I've updated the mockup above to include an initial section to represent the general idea. I think someone (perhaps @JohnMarkOckerbloom ?) probably should give some thought to exactly what we want to present here and how it relates to other data use related documentation, such as the Google doc you linked to. For example:

- Should that Google doc be converted to a page in the application wiki? That would feel more integrated and seems more sustainable/flexible than linking to a Google doc.
- Alternatively, the entire contents of that Google doc could be included in the new Data page, perhaps in a section collapsed by default but could be expanded by the user if they wanted to read the details.

JohnMarkOckerbloom:
[...]

PR #593, opened 2022-04-13
Issue517 create data page

Hurlyburly2:
Adds data page as outlined in https://github.com/pod4lib/aggregator/issues/517.
Let me know if there are any differences in ways that this would've been implemented, happy to accept feedback.

Screen Shot 2022-04-13 at 5 56 21 PM


ggeisler:
@Hurlyburly2 I'd suggest removing the Bootstrap classes on the heading elements (e.g. the `.mt-5 .mb-4` classes on `<h2 class="mt-5 mb-4">`). To ensure consistency throughout the application, we should rely on the default styling that comes with the different heading levels. If for some reason, we want to adjust them, I think it's better to do that in a SCSS file where we can apply the adjusted styling to the heading level globally, throughout the application.

I'm happy to go through the application and adjust visual design things like heading spacing later, when the application is more fully updated.

config/routes.rb
```diff
@@ -25,6 +25,8 @@
   get 'dashboard/uploads', to: 'dashboard#uploads', as: :activity
   get 'charts/uploads', to: 'charts#uploads', as: :uploads_chart
   get 'charts/records', to: 'charts#records', as: :records_chart
+  
+  get '/data', to: 'data#index'
```
marlo-longley, line 29: @Hurlyburly2 this is more a question since I'm newer to rails - what is the reason not to do resourceful routing since index is a default/included route? something like `resources :data, only: [:index]`

Hurlyburly2:
@ggeisler I've made the requested class removals- it does make the page look a little smooshed together, but as long as a style pass is planned then that's no problem.
I also added some additional tests regarding page access, which revealed that only admin had access to the /data route. I've refactored so logged in users also have access to the data page. While it's behaving as desired, I'm unsure if the `skip_authorization_check` usage is correct in the Data Controller. Please keep an eye on that in particular when you next take a look.

config/routes.rb
Hurlyburly2: @marlo-longley I was unsure of the actual differences myself (my Rails is kind of rusty!) so I checked with a coworker who's much more expert. Both methods create the same route, but `resources` establishes some helper functions and corresponds to a resource. In this case, we're serving up a pretty static page, and there is no model for Data to which CRUD stuff is being done.

ggeisler:
> I've made the requested class removals- it does make the page look a little smooshed together, but as long as a style pass is planned then that's no problem.

Sounds good @Hurlyburly2. I'll definitely be doing a thorough round of visual design updates. If the page formatting looks really bad, I'll do that sooner rather than later.

app/controllers/data_controller.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
```
Hurlyburly2, line 1: use existing pages controller

app/controllers/data_controller.rb
```diff
@@ -0,0 +1,8 @@
...
+
+# :nodoc:
+class DataController < ApplicationController
```
corylown, line 4: Instead of creating a new controller for this, I think using the existing `PagesController` would be appropriate because it's a static content page. Just add a new action `def data; end` to the pages controller.

corylown: Authorization in the controller action could look like:

```
  def data
    authorize! :read, :pages_data
  end
```

app/views/data/index.html.erb
```diff
@@ -0,0 +1,34 @@
+<div class="container index-page">
```
corylown, line 1: You would move and rename this file to `app/views/pages/data.html.erb`

app/views/shared/_nav.html.erb
```diff
@@ -12,7 +12,8 @@
       <ul class="navbar-nav ms-md-5">
         <li class='nav-item'><%= link_to t('.providers'), organizations_url, class: 'nav-link' %></li>
         <% if can? :manage, :dashboard_controller %>
-        <li class='nav-item'><%= link_to t('.activity'), activity_url, class: 'nav-link', data: { turbolinks: false } %></li>
+          <li class='nav-item'><%= link_to t('.activity'), activity_url, class: 'nav-link', data: { turbolinks: false } %></li>
+          <li class='nav-item'><%= link_to t('.data'), data_url, class: 'nav-link', data: { turbolinks: false } %></li>
         <% end %>
```
corylown, line 17: Wrap the data link in its own ability condition:

```suggestion
          <li class='nav-item'><%= link_to t('.activity'), activity_url, class: 'nav-link', data: { turbolinks: false } %></li>
        <% end %>
        <% if can? :read, :pages_data %>
          <li class='nav-item'><%= link_to t('.data'), data_url, class: 'nav-link', data: { turbolinks: false } %></li>
        <% end %>
```

config/locales/en.yml
```diff
@@ -57,6 +57,25 @@ en:
         contributing: Help with contributing or using POD data
         framework: Data provider and usage framework
         github: GitHub repository
+  data:
+    index:
```
corylown, line 61:
```suggestion
  pages:
    data:
```

config/routes.rb
corylown, line 29:
```suggestion
  get '/data', to: 'pages#data'
```

corylown (changes requested):
In addition to the inline comments/changes you'll need to update the ability model, something like:

```
    if user.roles.any?
      # ...
      can :read, :pages_data
    end
```

corylown (approved):

PR #594, opened 2022-04-13
Redesign streams header

### LINKED ISSUE #529: Update layout of stream title and actions
Related to #527.

provider-page-updates 2

---

**A: Stream name, date, default status** 
- [x] Replace the current "Name:" label with `Stream:`
- [x] Change the current `h5` class on this element to `h3`
- [x] If the stream is the provider's default stream, display a `badge-info` right after the stream name

**B: Streams link**
- [x] Rename the current "Streams" link to `Manage streams`
- [x] Remove the link from the `h2` element
- [x] Re-position the link to directly follow the stream name, or default badge, if present

**C: Stream actions**
- [x] Re-position the "Upload file" and "Re-analyze" buttons to align-right, in the same row as the stream file name and related info. There is currently a "Create new stream" button in roughly that location but...
- [x] ...Remove the "Create new stream" button. I think this action should be done through the Manage streams page (presumably it is not used frequently and will help reduce clutter on the main provider page).

marlo-longley:
Fixes #529 

Screen Shot 2022-04-13 at 15 42 51
Screen Shot 2022-04-13 at 15 42 28

### DIFF
--- a/app/views/shared/_organization_header.html.erb
+++ b/app/views/shared/_organization_header.html.erb
@@ -7,7 +7,7 @@
           <% if @organization.icon.attached? %>
             <%= image_tag(@organization.icon, class: 'organization-icon') %>
           <% end %>
-          <%= link_to_unless controller_name == 'organizations', @organization.name, @organization, class: 'text-body' %>
+          <%= @organization.name %>
         </h1>
 
         <p class="lead p-2">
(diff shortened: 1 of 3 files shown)


ggeisler:
@marlo-longley I don't know if you want to deal with this in this PR (I realize I didn't mention this in the ticketed design) or we can deal with it separately later, but I think we should remove the link on the organization name in the banner. With the Default stream link on the right size of the banner the user always has a way back to the main organization page, effectively duplicating the link on the organization name. It's not the most clear navigation, I realize, but I don't think it's good for us to offer two links in different spots that do the same thing.

marlo-longley:
@ggeisler I added a commit for removing the link :)

corylown (approved):

Issue #618, opened 2022-04-15
Can we produce statically generated oai responses like we do for normalized MARC dumps?


corylown:
This is @cbeer 's idea and I think it's worth considering:

- It would allow us to use processing jobs we already have in place to produce static oai response files. Performance should be good because we do the work in the background and just return a file on request.
- We'd be supporting "from date" requests in increments of 1 day since that's how frequently we produce deltas.
- To support "from date" requests that span multiple streams we could produce inter-stream delta files and return these instead of the full dump files.

Some issues to consider and questions:

- We would potentially be returning a large number of records in a single file (millions). This might be unexpected by some clients. But we're wondering if index data could handle this on an initial request for a full dump. Incremental deltas would be smaller.
- Is it a problem if we have a stream that has deltas but no full dumps? I think I've seen this scenario and it could make returning inter-stream deltas a problem because we wouldn't know that the first deltas were really full dumps.

corylown:
@JohnMarkOckerbloom interested in your thoughts on this approach when you get a chance.

JohnMarkOckerbloom:
This seems like a potentially useful way to go.  I've also used static files for handling some requests (including possibly some OAI requests) for The Online Books Page.  As I implement it, the files aren't literally the response, but the literal response either puts a wrapper around the files or does a trivial transformation on them in the output.  Either way, the response is quite fast.

One-day granularity is explicitly mentioned in the OAI spec as reasonable.  You can indicate in your response to the Identify verb that that's what you provide.

Index Data might or might not be able to handle millions of records in a single response.  (Even if they can, other potential consumers might have trouble with it.)  The usual "etiquette" I've seen in OAI-PMH providers is to break up large responses into chunks (where chunks typically range from 100-1000 records).  This is often handled server-side by first preparing the full response data (which would be what's in the static files in this scenario), and then returning the requisite number of records in the first response, along with a resumption token that, when included in a later request, essentially acts like a cursor telling the server where to start further responses.  This should hopefully not be too hard to implement as a transformation if need be.

JohnMarkOckerbloom:
As I've noted in other tickets today, it sounds like Index Data needs to have responses in limited-size chunks (of no more than a few thousands records per chunk).  This is not necessarily incompatible with statically generated OAI responses, but those responses, if created, may need to span multiple files, and/or be dynamically generated from appropriate portions of the larger static response.

corylown:
Closing this. This is the approach we're currently working toward.

PR #619, opened 2022-04-15
Spike at producing static OAI-PMH ListRecords responses with the existing dumps

cbeer:
See #618

### DIFF
--- a/app/controllers/oai_controller.rb
+++ b/app/controllers/oai_controller.rb
@@ -1,0 +1,117 @@
+# frozen_string_literal: true
+
+# Produce OAI-PMH responses
+class OaiController < ApplicationController
+  load_and_authorize_resource :organization
+
+  before_action do
+    @stream = @organization.default_stream
+    authorize! :read, @stream
+  end
+
+  def show
+    if params[:verb] == 'ListRecords'
+      render_list_records
+    else
+      render_error('badVerb')
+    end
+  end
+
+  private
+
+  def render_list_records
+    headers['Cache-Control'] = 'no-cache'
+    headers['Last-Modified'] = Time.current.httpdate
+    headers['X-Accel-Buffering'] = 'no'
+
+    self.response_body = build_list_records_response(*list_record_normalized_dump_candidates.first(2))
+  end
+
+  def render_error(code)
+    error = <<-XML
+      <?xml version="1.0" encoding="UTF-8"?>
+      <OAI-PMH xmlns="http://www.openarchives.org/OAI/2.0/"
+              xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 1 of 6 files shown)


app/controllers/oai_controller.rb
```diff
@@ -0,0 +1,117 @@
...
+  # rubocop:disable Metrics/AbcSize
+  def list_record_normalized_dump_candidates
+    @list_record_normalized_dump_candidates ||= begin
+      current_full_dump = @stream.current_full_dump
+
+      dumps = @stream.normalized_dumps.where(id: current_full_dump.id).or(current_full_dump.deltas).order(:created_at)
+
+      dumps = dumps.where(created_at: (Time.zone.parse(params[:from]).beginning_of_day)...) if params[:from]
+      dumps = dumps.where(created_at: ...(Time.zone.parse(params[:until]).end_of_day)) if params[:until]
+      dumps = dumps.where(id: params[:resumptionToken]...) if params[:resumptionToken]
```
thatbudakguy, line 69: is it valid to specify all of these at once (or any, in conjunction with the `resumptionToken`?) guessing that if you did send any other params with a `resumptionToken` you might end up with an empty set here, which i guess is OK behavior if it's an invalid request

app/controllers/oai_controller.rb
```diff
@@ -0,0 +1,117 @@
...
+      yielder << <<-EOXML
+      <?xml version="1.0" encoding="UTF-8"?>
+      <OAI-PMH xmlns="http://www.openarchives.org/OAI/2.0/"
+              xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+              xsi:schemaLocation="http://www.openarchives.org/OAI/2.0/
+              http://www.openarchives.org/OAI/2.0/OAI-PMH.xsd">
+        <responseDate>#{Time.zone.now.strftime('%F')}</responseDate>
+        #{build_request.to_xml}
+        <ListRecords>
+      EOXML
```
thatbudakguy, line 88: what is `EOXML`? why is it used here vs. just `XML` in the error response?

app/controllers/oai_controller.rb
```diff
@@ -0,0 +1,117 @@
...
+      render_error('badVerb')
+    end
+  end
+
+  private
+
+  def render_list_records
+    headers['Cache-Control'] = 'no-cache'
+    headers['Last-Modified'] = Time.current.httpdate
+    headers['X-Accel-Buffering'] = 'no'
```
thatbudakguy, line 25: what does `'X-Accel-Buffering'` do? does it cause problems with the response?

thatbudakguy:
pre-emptively merging as per discussion with @corylown so that we have a starting point for other parts of the OAI implementation.

PR #620, opened 2022-04-15
Add tab navigation and update layouts for Provider page

### LINKED ISSUE #531: Set up tab navigation for main provider-related pages

### LINKED ISSUE #607: Reanalyze button takes user to confusingly similar page
I just want to make sure this is intended. Steps:

1. From the default organization page (here /organizations/stanford), click "Reanalyze"
Screen Shot 2022-04-14 at 14 12 20

2. You are routed to the stream's specific page at (/organizations/stanford/streams/thedefault) which looks like this:
Screen Shot 2022-04-14 at 14 12 32

The page is very similar, just without the tabs we are implementing for the Provider page. This confused me, I didn't know where I was in the app. Is this intended? Should the stream pages look different than the Provider Pages? Should they also implement the tabs? I am likely missing some context, thanks!

marlo-longley:
Fixes #531
Fixes #607

Also improves testing for existing organization views, and aligns permissions with our permissions doc.

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -49,7 +49,7 @@ class Ability
 
     can :manage, :all if user.has_role?(:admin)
     cannot :delete, Stream, default: true
-    can :read, Organization, public: true
+    can %i[read normalized_data], Organization, public: true
     can :manage, :dashboard_controller if user.has_role?(:admin)
 
     owned_orgs = Organization.with_role(:owner, user).pluck(:id)
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -44,8 +44,7 @@ class Ability
       can :read, ActiveStorage::Attachment, { record: { organization: { public: true } } }
       can :read, MarcRecord, upload: { organization: { public: true } }
       can %i[read profile info], [Stream, Upload], organization: { public: true }
-      # TODO: adding OVerciew model for clarity?
-      can :read, :pages_data, Overview
+      can :read, :pages_data
       can %i[read users organization_details provider_details normalized_data], Organization, public: true
     end
 
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -44,12 +44,14 @@ class Ability
       can :read, ActiveStorage::Attachment, { record: { organization: { public: true } } }
       can :read, MarcRecord, upload: { organization: { public: true } }
       can %i[read profile info], [Stream, Upload], organization: { public: true }
-      can :read, :pages_data
+      # TODO: adding OVerciew model for clarity?
+      can :read, :pages_data, Overview
+      can %i[read users organization_details provider_details normalized_data], Organization, public: true
     end
 
     can :manage, :all if user.has_role?(:admin)
     cannot :delete, Stream, default: true
-    can %i[read normalized_data], Organization, public: true
+    can :read, Organization, public: true
     can :manage, :dashboard_controller if user.has_role?(:admin)
 
     owned_orgs = Organization.with_role(:owner, user).pluck(:id)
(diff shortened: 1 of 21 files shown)


marlo-longley:
Member view:

Screen Shot 2022-04-15 at 15 25 37

Owner view: 

Screen Shot 2022-04-15 at 15 26 06

app/models/ability.rb
```diff
@@ -54,13 +54,15 @@ def initialize(user, token = nil)
     owned_orgs = Organization.with_role(:owner, user).pluck(:id)
     can :manage, Organization, id: owned_orgs
     can %i[crud profile info], [Stream, Upload], organization: { id: owned_orgs }
+    cannot %i[reanalyze], [Stream], organization: { id: owned_orgs }
```
corylown, line 57: I don't think this in needed. Org owners cannot re-analyze streams even if this line is not present because they can `create` and `read` streams, but don't have full `manage` permissions.

app/models/ability.rb
```diff
@@ -54,13 +54,15 @@ def initialize(user, token = nil)
...
     can :read, MarcRecord, upload: { organization: { id: owned_orgs } }
     can :manage, AllowlistedJwt, resource_type: 'Organization', resource_id: owned_orgs
     can :read, ActiveStorage::Attachment, { record: { organization: { id: owned_orgs } } }
 
     member_orgs = Organization.with_role(:member, user).pluck(:id)
-    can %i[invite read users organization_details provider_details], Organization, id: member_orgs
+    can %i[invite read users organization_details provider_details normalized_data], Organization, id: member_orgs
     can %i[read profile info], [Stream, Upload], organization: { id: member_orgs }
+    can %i[create], [Upload], organization: { id: member_orgs }
```
corylown, line 65: Not 100% sure what behavior we want?

marlo-longley: @corylown my reading of the permissions doc and comments in tickets was that members can create uploads, but not create streams. Did you have a different understanding?

marlo-longley: @corylown Hm just looked and I think this may have been changed in the doc. I can remove upload function from members?

corylown: @marlo-longley I guess leave what you have. I'm not really sure what we want to allow members to do.

app/models/ability.rb
marlo-longley: @corylown I removed this line.

corylown:
@marlo-longley a couple other things to check permissions-wise:

* As an org member I should be able to view other orgs' normalized files, but I cannot.
* As an org member I should NOT be able to view other orgs' token page, but I can. (The tokens don't appear but we should probably just hide the tab and the page if the user does not have permission to view them)

marlo-longley:
@corylown your points about permissions should be addressed in this new commit. Thanks for catching those!

https://github.com/pod4lib/aggregator/pull/620/commits/3633ce38a43bdd726a731c0dd13eb3e1d7cd46fa

corylown:
@marlo-longley I think this is still not quite right. Maybe we can pair on this when you're ready?

corylown (approved):

PR #623, opened 2022-04-18
Cannot have more than one pages translation key

corylown:


cbeer:
Should we consider adding https://rubygems.org/gems/i18n-tasks?

PR #627, opened 2022-04-18
Manage organization slug autofill and permissions

### LINKED ISSUE #566: Update how we poplulate and display the org "Slug" field
Extracted from discussion in #546. This is not a high-priority issue, but a nice-to-have when other organization-related updates such as #546 have been resolved.

The overall goal is to simplify creating a slug for a new POD organization and prevent it from being easily editable after initial creation (since we don't think it's something that is likely to need editing after creation).

- [ ] On the New organization form `/organizations/new`, populate the Slug input as the Name input is filled in, but enable the Slug input to remain editable. See Spotlight (create a new exhibit form) for an example (and there's a good chance @cbeer implemented that, if there are questions).
- [ ] On the Manage organization page, Organization details tab, only display the Slug label/input field if the user is a POD admin/superuser (that is, don't display on the page when the user is an org admin unless they have the higher powers). That way, in the rare case it needs to be updated, it can be, but only by a smaller set of people who presumably know why it needs to be updated. Others won't see the field and thus won't be confused by what is for or change it unnecessarily.
- [ ] When the user is an org member, and thus looking at the read-only version of this page as described in the second part of #546, I think we should omit display of the Slug label and value, since it doesn't seem necessary for users to really know this value for anything they do in POD.

ktamaral:
Fixes #566 

In the future, it would be best to move URLify to a reusable js module

### DIFF
--- a/app/javascript/organizations.js
+++ b/app/javascript/organizations.js
@@ -1,0 +1,148 @@
+window.addEventListener('load', function () {
+
+  let organization_name = this.document.getElementById('organization_name');
+  let organization_slug = this.document.getElementById('organization_slug');
+
+  organization_name.addEventListener("keyup", function (e) {
+    const nameVal = e.target.value;
+    organization_slug.setAttribute('placeholder', URLify(nameVal, nameVal.length));
+  });
+
+  organization_slug.addEventListener("focus", function (e) {
+    const slugVal = e.target.value;
+    if (slugVal === '') {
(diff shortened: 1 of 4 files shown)


marlo-longley (approved):
Looks good to me. The downcoding maps are a lot, but it makes sense to include! I can see this is how it's implemented in Spotlight as well.

Issue #545, opened 2022-04-04
Display the organization contact value as a link


ggeisler:
EDIT: we settled on only using a single email address for the organization contact field. I've updated some of the original ticket description to match that decision.

In #544 we're moving the organization email contact to be a single-valued attribute of organization. ~~We've also discussed allowing the value of this field to be either an email address or a URL (e.g., maybe the organization wants people to go to a page they've set up locally to provide more info about their POD stuff).~~

~~I'm not sure if there are any technical hurdles to storing either an email address or a URL in the same field, or in then displaying whichever is stored as a different type of link, but I'm assuming not and we can discuss alternatives if the idea below doesn't make sense.~~

- [x] Display the value of the organization contact field as a link in the organization banner, with the label `POD contact for this organization`:

<img width="1197" alt="Screen Shot 2022-04-04 at 4 11 51 PM" src="https://user-images.githubusercontent.com/101482/161647396-01d19f47-e1b3-4606-b696-c666f519f2e0.png">

---

- [x] ~~If the value is an email address,~~ use the `mailto:` form of the link, ~~otherwise a standard link~~
- [x] If the organization has not entered a value for their contact, don't display the label:

<img width="1197" alt="Screen Shot 2022-04-04 at 4 12 36 PM" src="https://user-images.githubusercontent.com/101482/161647505-28bdc8d1-b8f8-4dc6-9815-0b0bb3b40867.png">

ktamaral:
The contact field is a multi-valued email field right now. The proposal is to make it single-valued. We will have to deal with the edge case of multiple values until that gets settled. Also we are only expecting email for that field and not a url to a form so that simplifies things a bit.

JohnMarkOckerbloom:
I have not to date heard from anyone who needs a contact URL.  I think we can make this simply a single email address for now.

ggeisler:
Sounds good @JohnMarkOckerbloom.

We still have the single-valued versus multi-valued aspect of this to deal with. Although we only have one organization using more than one value, and I feel it would simplify things without being unreasonable to make this field single-valued and ask the one org to settle on a single email address for their contact, I suppose another option would be to maintain this field as multi-valued and use all values in a single mailto link. Here's an example of how that might be done (although you might be able to add all values to the "to" part, instead of using "cc", maybe?):

https://css-tricks.com/snippets/html/mailto-links/#aa-adding-cc-and-bcc

JohnMarkOckerbloom:
[...]

Issue #535, opened 2022-04-01
Adjust Manage/View organization page when org is consumer-only


ggeisler:
Related to #534.

Although most POD organizations are providers, we also expect to have at least one consumer-only organization. Some of the details described in #534 don't apply to consumer-only orgs, since they don't have access tokens or MARC details, but others do. 

The details below assume #534 is completed first, so these are the differences from #534 when the organization is consumer-only:

<img width="1198" alt="Screen Shot 2022-04-04 at 12 26 27 PM" src="https://user-images.githubusercontent.com/101482/161617517-f218c69c-e34f-4405-a27f-ef0d53445e91.png">

---

- [x] **A:** Since a consumer-only org doesn't have streams, we should omit the "Default stream" link we show to provider organizations.
- [x] **B:** Since a consumer-only org doesn't have access tokens or MARC details, we should omit the "Access tokens" and "Provider details" tabs we show to provider organizations.

The form on the Organization details page might need to be slightly different as well, because consumer-only orgs don't have a MARC organization code but I'll address that in a separate ticket about that particular page.

thatbudakguy:
@ggeisler i'm getting confused about what the consumer org page should display (although maybe we don't have the answer to that yet).

- for providers, their homepage (what you get if you go to `/organizations/my-org`) displays their default stream and other relevant data. obviously this doesn't work for a consumer org.
- for providers, you get a link up in the top right to go to the org users page `/organizations/my-org-users`, which looks different depending on whether you are an owner or not. from here, you can get to any of the other "manage/view organization" pages, which we have styled to appear as tabs.

what should display when you go to the homepage for a consumer org? i.e. `/organizations/my-org`? right now, there's not really anything valid we could display, because it's all geared toward streams. i can't tell if the designs are implying that the content from the manage/view organization pages should be shown _on the org homepage_ for a consumer, but i don't think they are. should we just leave it blank?

hope this makes sense; not directly related to this issue but i figured i should raise it regardless.

thatbudakguy:
other random thought...there are currently 3 ways to get to a consumer org's pages:
- click "providers" in the nav, which gives you a list of all organizations, not just providers (probably should fix this)
- log in as a member of a consumer org and visit the homepage, which gives you a link to your org
- directly paste/type in a link to a consumer org's page

and similarly, there's only 2 ways to create a consumer org:
- click "providers" in the nav and then click "create organization" and uncheck the "provider" box
- do it manually in the rails console

[...]

PR #647, opened 2022-04-19
Organization page updates for consumers

thatbudakguy:
This PR does a few small, related things:

- Remove stream display from consumer org homepage (there's nothing to display here, at present)
- Update organization view/manage pages for consumer orgs (#535)

It also updates the org header layout slightly to remove unneeded CSS and align everything depending on what content is available. e.g. for a provider org with logo, MARC code, and contact email:

Screen Shot 2022-04-19 at 15 09 18

but for a consumer org with no logo and no MARC code:

Screen Shot 2022-04-19 at 15 09 36

(flexbox highlighted to show the alignment logic).

the contact email logic assumes `Organization.contact_emails` will still exist; that's easy enough to change with #544 if we do so. this PR thus fixes/refs #545.

### DIFF
--- a/app/views/shared/_manage_organization_header.html.erb
+++ b/app/views/shared/_manage_organization_header.html.erb
@@ -10,13 +10,12 @@
 <div class="d-flex flex-row justify-content-center">
     <div class="btn-group manage-organization-tabs" role="group" aria-label="Manage organization tabs">
         <%= link_to "Users", organization_users_path(@organization), class: "btn btn-outline-primary #{current_page_class(organization_users_path(@organization))}" %>
-
-        <% if can? :read, @organization.allowlisted_jwts.build %>
-          <%= link_to "Access tokens", organization_allowlisted_jwts_path(@organization), class: "btn btn-outline-primary #{current_page_class(organization_allowlisted_jwts_path(@organization))}" %>
+        <% if @organization.provider? && can? :read, @organization.allowlisted_jwts.build %>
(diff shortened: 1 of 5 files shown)


app/views/shared/_manage_organization_header.html.erb
```diff
@@ -10,13 +10,12 @@
 <div class="d-flex flex-row justify-content-center">
     <div class="btn-group manage-organization-tabs" role="group" aria-label="Manage organization tabs">
         <%= link_to "Users", organization_users_path(@organization), class: "btn btn-outline-primary #{current_page_class(organization_users_path(@organization))}" %>
-
-        <% if can? :read, @organization.allowlisted_jwts.build %>
-          <%= link_to "Access tokens", organization_allowlisted_jwts_path(@organization), class: "btn btn-outline-primary #{current_page_class(organization_allowlisted_jwts_path(@organization))}" %>
+        <% if @organization.provider? && can?(:read, @organization.allowlisted_jwts.build) %>
```
corylown, line 13: Consumer orgs need to be able to manage tokens so they can consume data via the API.
```suggestion
        <% if can? :read, @organization.allowlisted_jwts.build %>
```

thatbudakguy: thanks; I was unsure about this — @ggeisler seemed to think otherwise in https://github.com/pod4lib/aggregator/issues/535.

ggeisler: Well, I **assumed** consumers didn't need to do anything with tokens, obviously not understanding what tokens are for (though I guess I'm starting to understand..).

corylown (approved):

PR #655, opened 2022-04-20
Update homepage links

marlo-longley:
We have some new pages and tabs, so align the homepage links with them.


thatbudakguy (approved):
thanks for fixing these!

Issue #426, opened 2022-02-03
Right-align numeric table columns


ggeisler:
The various tables currently in POD tend to contain a lot of numeric data. Numeric table data is easier to scan, comprehend, and compare (e.g., larger values are more apparent at a glance) when they are right-aligned in the table column.

We'll probably want to do other visual refinement related to POD tables, but a good start would be to right-align the numeric columns listed in the tables below. Note that for each column where the values are right-aligned, the table header for that column should also be right-aligned.

(It would be nice to right-align the Size column in relevant tables as well, since this is basically a numeric field. But we're including units in the value of that column, and several different units at that ("GB" versus "Bytes"), which complicates things. So maybe we can leave this column as-is for now and come back to it when other table improvements have been completed.)

Headings below are table names; checkboxes are the columns that need realignment.

#### Organizations
- [x] Files
- [x] Unique records
- [x] Total records

#### Organization
- [x] Records

#### Normalized data
- [x] Records

#### Activity/Uploads
- [x] File count
- [x] Records

marlo-longley:
@ggeisler I'm starting to look into this, thanks for compiling it.

I noticed that in the MVP codebase, it looks like the intention was to have every column including headers and non-numeric columns have the text align to the right. Maybe this wasn't an intention, just a random thing. But you can see the `text-right` class applied to everything. However, that class was removed in BS5, so that's why everything reverted back to left alignment.
![Screen Shot 2022-03-17 at 15 28 45](https://user-images.githubusercontent.com/1328900/158904840-d1bb35e4-b526-43e1-a86e-9cae59f3341b.png)

I'm wondering if I should realign everything to the right again, rather than pick and chose only the numeric columns like you indicated here.

ggeisler:
@marlo-longley Sure, if you want to just change all the `text-right` classes to `text-end` we can see how that looks and go from there. You're right, that might be a simpler way to go given the current code.

marlo-longley:
@ggeisler awesome, thanks for your input! Seems simple for now.

marlo-longley:
Update: deciding to wait on this until more clarity on which tables and views will make it into the next design of the app.

cbaksik:
Having commas in the numbers is very helpful as well. I see them now but don't think they were always there. Wanted to affirm that they are desired.

ggeisler:
👍 I'll keep an eye on that as we finalize the display details later in the workcycle.

corylown:
Closed by #656

[...]

PR #656, opened 2022-04-20
Harmonize table styles

thatbudakguy:
this PR makes a small number of changes to table displays so that they're more consistent.

* uses `align-middle` so that table cell contents are always vertically centered
* replaces `text-left` and `text-right` with BS5 `text-start` and `text-end` so that text is correctly aligned
* tries to better align a few table headers with their content
* makes buttons inside table rows all `.btn-sm` so they impact layout less and are consistent

### DIFF
--- a/app/assets/stylesheets/organizations.scss
+++ b/app/assets/stylesheets/organizations.scss
@@ -9,7 +9,7 @@
 // Table of all organizations
 .organizations {
   .icon-column {
-    min-width: 68px; // Try to avoid some layout shifting when icons are rendered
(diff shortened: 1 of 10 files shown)


marlo-longley:
@thatbudakguy would you say this addresses #426 ?

thatbudakguy:
oh I hadn't even seen that issue. yes, probably? I'll confirm (and maybe check on the commas thing)

thatbudakguy:
ok, can confirm this closes #426.

marlo-longley (approved):

corylown (approved):

PR #662, opened 2022-04-20
Redirect missing consumer org pages

thatbudakguy:
Some pages don't make sense to display for consumer orgs, so this just adds some redirects to point people to the pages that are meaningful for them. See conversation on #535 for more info.

* Redirect consumer org home page to org details page
* Redirect consumer org provider details to org details page


marlo-longley:
@thatbudakguy the code looks fine to me, I know redirects should be used sparingly but seems like the best way without deeply rethinking a ton of logic. I read the conversation with you and Gary on the ticket.

marlo-longley (approved):

PR #665, opened 2022-04-21
Restrict access to tokens tab and route

### LINKED ISSUE #641: Properly restrict privileges for Access tokens
1. Only allow authorized users (POD admins, and owners of that specific org) to see the Access tokens tab in Manage Organization.

2. Not only do we want to hide the tab, but restrict the route itself at /organizations/:org/allowlisted_jwts. Currently unprivileged users can navigate to this route, even though the tokens themselves are not displayed.

Screen Shot 2022-04-19 at 11 58 02

ktamaral:
Fixes #641

### DIFF
--- a/app/controllers/allowlisted_jwts_controller.rb
+++ b/app/controllers/allowlisted_jwts_controller.rb
@@ -4,6 +4,7 @@
 class AllowlistedJwtsController < ApplicationController
   load_and_authorize_resource :organization
   load_and_authorize_resource through: :organization
+  authorize_resource :class => :controller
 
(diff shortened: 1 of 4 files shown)


marlo-longley:
@ktamaral seems good, I pulled and it works well. I am new to this as well so I read the docs here: https://github.com/ryanb/cancan/wiki/Non-RESTful-Controllers - 11 year old docs so maybe you read something else!

I found the syntax of `authorize_resource class: false`  as opposed to `authorize_resource class: :controller` to be more readablebut they both work the same I guess? Anyway happy to approve.

PR #667, opened 2022-04-21
Contact email refactor

thatbudakguy:
- Contact email validation
- Add contact email to organization page
- Update org contact email relationship and routes
- Remove unused contact email functionality
- Allow creating contact email with organization
- Refactor org details and provider details forms
- Remove unused test
- Render warning when org details can't be updated


app/controllers/organizations_controller.rb
```diff
@@ -67,7 +67,7 @@ def update
         format.html { redirect_to @organization, notice: 'Organization was successfully updated.' }
         format.json { render :show, status: :ok, location: @organization }
       else
-        format.html { render :edit }
+        format.html { redirect_to @organization, alert: 'Organization could not be updated.' }
```
thatbudakguy, line 70: feel like we should do something more helpful here, like print the error, but I tried `@organization.errors` and there didn't appear to be anything there. this can occur if you try to remove a contact email, since we require one.

app/models/organization.rb
```diff
@@ -12,9 +12,10 @@ class Organization < ApplicationRecord
   has_many :uploads, through: :streams
   has_many :marc_records, through: :streams, inverse_of: :organization
   has_many :allowlisted_jwts, as: :resource, dependent: :delete_all
-  has_many :contact_emails, dependent: :delete_all
+  has_one :contact_email, dependent: :delete
   has_one_attached :icon
   has_many :statistics, dependent: :delete_all, as: :resource
+  accepts_nested_attributes_for :contact_email, update_only: true
```
thatbudakguy, line 18: the result of this is that:
* if you create a new org, you must specify a single contact email
* if you edit the org details, you can change the email, but you can't remove it
* existing orgs without a contact email stay that way, until you add one

which i think is the behavior we want.

app/views/organizations/organization_details.html.erb
```diff
@@ -1,63 +1,47 @@
...
+    </dt>
+    <dt>
+      <h4>Code</h4>
+      <p class="<%= @organization.code.blank? ? 'fst-italic' : '' %>">
+        <%= @organization.code.blank? ? 'no value provided' : @organization.code %>
+      </p>
+    </dt>
+    <dt>
+      <h4>POD contact for this organization</h4>
+      <p>TBD after contact field is added</p>
```
corylown, line 36:
```suggestion
      <p class="<%= @organization.contact_email.blank? ? 'fst-italic' : '' %>">
        <%= @organization.contact_email.blank? ? 'no value provided' : @organization.contact_email.email %>
      </p>
```

app/models/organization.rb
corylown: I'm a bit wary of making this field a hard requirement to create an organization, but we can address that question as a separate issue.

corylown (approved):

PR #668, opened 2022-04-22
Prefer resourceful abilities

cbeer:
@ktamaral @marlo-longley In #665, I see you added this non-resourceful ability check, but I'm not sure  of its purpose when we already have the resource-oriented ability defined:

```
can :manage, AllowlistedJwt, resource_type: 'Organization', resource_id: owned_orgs
```

Can you clarify the purpose of the controller authorization, or make sure to add a test for it?

### DIFF
--- a/app/controllers/allowlisted_jwts_controller.rb
+++ b/app/controllers/allowlisted_jwts_controller.rb
@@ -4,7 +4,6 @@
 class AllowlistedJwtsController < ApplicationController
   load_and_authorize_resource :organization
   load_and_authorize_resource through: :organization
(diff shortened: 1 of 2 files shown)


marlo-longley:
@cbeer thanks for looking this over Chris. The purpose of this PR was to restrict access to the route `/organizations/:org/allowlisted_jwts`. Prior to this change unprivileged users could navigate to the route even though once they got there, the tokens were restricted. Definitely want to know the best practice for this, though.

corylown:
I reverted the original change via #682

PR #686, opened 2022-04-25
Restyle turbolinks progress bar

### LINKED ISSUE #658: Add an indication that the MARC analysis page is loading
From the main provider page, when you select the "MARC analysis" tab, that page takes a while to load. But from a user perspective it isn't necessarily clear what is happening; a user could think the tab just isn't working.

We should add some sort of indication of the page loading situation.

thatbudakguy:
Fixes #658


thatbudakguy:
I ended up really liking the constrasting yellow look. Changed the behavior a bit so it doesn't shift the nav down when it appears, which was causing a vertical scrollbar to briefly appear.

corylown (approved):

PR #687, opened 2022-04-25
Update mailer text

### LINKED ISSUE #639: Update user invitation mailer text
This is not a high-priority (the current mailer works and is understandable) but I think we could make some minor improvements to the email that is sent when a user is invited to a POD organization.

- [ ] Subject line: `POD Aggregator invitation instructions` replaces current subject line string
- [ ] Remove link from the invitee's email address in the "Hello" line. 
- [ ] Update the paragraph that currently begins with "Someone has invited you..." to `You've been invited to join the [organization name] organization in the POD Aggregator (https://pod.stanford.edu). You can accept this invitation using the link below.` where [organization name] is replaced with the name of the org that invited the user (e.g. "Stanford"). Maintain the link for the POD url as it currently is.

If it is technically difficult to get the org name in this context, we could go with `You've been invited to join the POD Aggregator (https://pod.stanford.edu). You can accept this invitation using the link below.` instead (but including the org name would be nice because it helps legitimize the email, hopefully).

ktamaral:
Fixes #639 (see comments in ticket)

### DIFF
--- a/app/views/devise/mailer/invitation_instructions.html.erb
+++ b/app/views/devise/mailer/invitation_instructions.html.erb
@@ -1,6 +1,6 @@
 <p><%= t("devise.mailer.invitation_instructions.hello", email: @resource.email) %></p>
 
-<p><%= t("devise.mailer.invitation_instructions.description", url: root_url) %></p>
+<p><%= t("devise.mailer.invitation_instructions.someone_invited_you", url: root_url) %></p>
 
 <p><%= link_to t("devise.mailer.invitation_instructions.accept"), accept_invitation_url(@resource, invitation_token: @token) %></p>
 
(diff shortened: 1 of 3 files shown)


app/views/devise/mailer/invitation_instructions.html.erb
```diff
@@ -1,6 +1,6 @@
 <p><%= t("devise.mailer.invitation_instructions.hello", email: @resource.email) %></p>
 
-<p><%= t("devise.mailer.invitation_instructions.someone_invited_you", url: root_url) %></p>
+<p><%= t("devise.mailer.invitation_instructions.description", url: root_url) %></p>
```
corylown, line 3: This is minor, but it might be preferable to continue using the `someone_invited_you` translation key because it's an override of a key coming from the gem `devise_invitable`.

corylown (approved):

Issue #638, opened 2022-04-19
A user should be either an org owner or org member not both.


corylown:
Should fix issues like: https://github.com/pod4lib/aggregator/issues/637

corylown:
If changing the underlying behavior of the roles proves to be too difficult or problematic it's also possible to change the logic for displaying users, something like:

```
# All users with the role :owner
owners = org.roles.select { |role| role.name == 'owner'  }.flat_map(&:users)

# All users with the role :member who are not also :owner
members = org.roles.select { |role| role.name == 'member'  }.flat_map(&:users).reject { |user| owners.include?(user) }
```

cbeer:
Rolify suggests adding this kind of association:

```
has_many :users, -> { distinct }, through: :roles, class_name: 'User', source: :users
```
https://github.com/RolifyCommunity/rolify/wiki/Usage#finding-roles-through-associations

🤷‍♂️

corylown:
Closed via #700

PR #700, opened 2022-04-26
Fix duplication in display of users and organizations 

marlo-longley:


corylown:
@marlo-longley I think Chris's comment was leading in the correct direction for a fix. You can add the following to the Organization model:

`  has_many :users, -> { distinct }, through: :roles, class_name: 'User', source: :users`

Then you can get each distinct user through an organization:

`Organization.where(slug: 'stanford').first.users`

And each of their roles:

`Organization.where(slug: 'stanford').first.users.first.roles`

Then the views can be updated to use this method of fetching users and roles to avoid display duplication.

marlo-longley:
@corylown thanks for looking Cory. I wanted to change the underlying logic so that users could not have both the owner and member role at once. Your solution just changes the view. Is this acceptable?

corylown:
I'm not convinced that a user having more than one role is problematic other than the display issue. This fixes the model so we can display users how people expect.

marlo-longley:
@corylown ok thanks for clarifying! I think I was reading the ticket title too literally. Will update soon.

marlo-longley:
Fixes #638 
Not sure if there was an issue for the navbar duplication too.

corylown (approved):

PR #701, opened 2022-04-26
Upload ordering

thatbudakguy:
- Add default ordering to Uploads
- Remove updated by ordering on Uploads dashboard

### DIFF
--- a/app/models/upload.rb
+++ b/app/models/upload.rb
@@ -14,6 +14,7 @@ class Upload < ApplicationRecord
   validate :valid_url, if: proc { |upload| upload.url.present? }
   scope :active, -> { where(status: %w[active processed]) }
   scope :archived, -> { where(status: 'archived') }
+  default_scope { order(created_at: :desc) }
 
   after_create :attach_file_from_url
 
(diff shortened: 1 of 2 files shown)


app/models/upload.rb
```diff
@@ -14,6 +14,7 @@ class Upload < ApplicationRecord
   validate :valid_url, if: proc { |upload| upload.url.present? }
   scope :active, -> { where(status: %w[active processed]) }
   scope :archived, -> { where(status: 'archived') }
+  default_scope { order(created_at: :desc) }
```
corylown, line 17: @thatbudakguy it looks like changing the default ordering has some unintended side-effects, like breaking the activity charts on the providers page (https://pod.stanford.edu/organizations):
Screen Shot 2022-04-27 at 8 05 30 AM

There may be other places in the application where changing the default order has unintended side-effects. I'm wondering if it would make sense to go back to setting the order in the `uploads` action instead?

corylown (changes requested):

app/models/upload.rb
thatbudakguy: dang, I wish we'd had tests for that! It does beg the question of why the charts are relying on `updated_at`, and whether they should be. i'll dig into this more

thatbudakguy: reverted the default ordering so i can focus on other stuff.

corylown (approved):

PR #702, opened 2022-04-26
Change link name and add link to org header

### LINKED ISSUE #692: There's no link back to organization home/default stream when viewing streams
The organization name is no longer linked and there's no link to the right to get me back where I came from:

https://pod.stanford.edu/organizations/penn/streams

Screen Shot 2022-04-25 at 4 50 48 PM

I would expect to find a link back to: https://pod.stanford.edu/organizations/penn

marlo-longley:
Fixes #692

### DIFF
--- a/spec/views/shared/layout_manage_organization.html.erb_spec.rb
+++ b/spec/views/shared/layout_manage_organization.html.erb_spec.rb
@@ -15,19 +15,19 @@ RSpec.describe 'shared/_layout_manage_organization', type: :view do
     assign(:stream, stream)
   end
 
-  it 'links to Default stream for provider orgs' do
+  it 'links to Provider home for provider orgs' do
     render
 
-    expect(view.content_for(:org_header)).to have_link 'Default stream'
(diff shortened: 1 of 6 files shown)


spec/views/shared/layout_manage_organization.html.erb_spec.rb
```diff
@@ -15,19 +15,19 @@
...
 
-    expect(view.content_for(:org_header)).to have_link 'Default stream'
+    expect(view.content_for(:org_header)).to have_link 'Provider home'
   end
 
   context 'with a consumer org' do
     let(:is_provider) { false }
 
-    it 'does not link to Default stream for consumer orgs' do
+    it 'does not link toProvider home for consumer orgs' do
```
corylown, line 27: Tiny typo:
```suggestion
    it 'does not link to Provider home for consumer orgs' do
```

corylown (changes requested):

marlo-longley:
@corylown typo should be fixed, thanks!

corylown (approved):

PR #714, opened 2022-04-27
Calculate interstream deltas

Hurlyburly2:
Creates a new job that, provided a stream (or stream id for use in manually running the job) will generate the deltas between its full dump and the last good state of the previous default stream in the default stream history (Last good state of previous stream = the full dump + deltas)

### DIFF
--- a/app/models/stream.rb
+++ b/app/models/stream.rb
@@ -13,7 +13,6 @@ class Stream < ApplicationRecord
   has_one :statistic, dependent: :delete, as: :resource
   has_many :normalized_dumps, dependent: :destroy
   has_many :job_trackers, dependent: :delete_all, as: :reports_on
-  has_many :interstream_deltas, dependent: :destroy
 
   scope :default, -> { where(default: true) }
   scope :active, -> { where(status: 'active') }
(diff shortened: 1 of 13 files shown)


app/models/stream.rb
```diff
@@ -7,12 +7,13 @@ class Stream < ApplicationRecord
...
   belongs_to :organization
   has_many :uploads, dependent: :destroy
-  has_many :default_stream_histories, dependent: :destroy
+  has_one :default_stream_history, dependent: :destroy
   has_many :marc_records, through: :uploads, inverse_of: :stream
   has_many :files, source: :files_blobs, through: :uploads
   has_one :statistic, dependent: :delete, as: :resource
   has_many :normalized_dumps, dependent: :destroy
   has_many :job_trackers, dependent: :delete_all, as: :reports_on
+  has_many :interstream_deltas, dependent: :destroy
```
corylown, line 16: I think this association is not needed (and may cause problems when an organization is deleted). Organizations already have a relationship to normalized_dumps and normalized_dumps have a relationship to interstream_deltas, so there is already a relationship between organizations & interstream_deltas through normalized_dumps.

corylown (approved):

PR #716, opened 2022-04-27
Fix typeo on manage org page

### LINKED ISSUE #715: Type-o on manage organization page
There seems to be a random Beta character on the manage organization page, likely a type-o

https://github.com/pod4lib/aggregator/blob/main/app/views/shared/_layout_manage_organization.html.erb#L8

POD_ManageOrganization_BetaCharacter

ktamaral:
Fixes #715


corylown (approved):
Good catch!

PR #733, opened 2022-04-28
OAI ListSets response

thatbudakguy:
This PR also moves the OAI-PMH api to `/oai` (instead of being per-organization), since [the spec says](https://www.openarchives.org/OAI/openarchivesprotocol.html#HTTPRequestFormat):
> There is a single base URL for all requests. The base URL specifies the Internet host and port, and optionally a path, of an HTTP server acting as a repository.

### DIFF
--- a/app/controllers/oai_controller.rb
+++ b/app/controllers/oai_controller.rb
@@ -4,11 +4,6 @@
 class OaiController < ApplicationController
   load_and_authorize_resource :organization
 
-  before_action do
-    @stream = @organization.default_stream
-    authorize! :read, @stream
-  end
-
   def show
     if params[:verb] == 'ListRecords'
       render_list_records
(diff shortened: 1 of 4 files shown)


app/controllers/oai_controller.rb
```diff
@@ -1,34 +1,47 @@
...
+# rubocop:disable Metrics/ClassLength
 class OaiController < ApplicationController
   load_and_authorize_resource :organization
 
-  before_action do
-    @stream = @organization.default_stream
-    authorize! :read, @stream
-  end
-
   def show
```
thatbudakguy, line 8: maybe this should be replaced with something, I'm just unfamiliar with how authorization works in controllers. should any authenticated user be able to `ListSets`?

corylown: It seems ok to allow any authenticated user to make a ListSets request.

corylown (approved):

Issue #634, opened 2022-04-19
Careful review of permissions in the application


corylown:
Review https://docs.google.com/spreadsheets/d/1_p8FkN_E_x5T_W1HF5bdX7m0eDt6xrLLMG8IIECBkcI/edit#gid=1891964460

And confirm behavior of the application conforms to our expectations.

marlo-longley:
- [change from before] Org members should be able to view their own access tokens, but not the tokens of other orgs

marlo-longley:
Closing -- we've reviewed permissions for current state of the app. Other things may come up which we will individually ticket.

PR #739, opened 2022-04-28
Add tests for verified permission behavior in app

marlo-longley:
This work is related to #634

I added tests for non-controversial permissions items in the spreadsheet that were missing tests. Everything tested here was in sync between the app and the spreadsheet. Getting these tests in before making further changes would be good.

### DIFF
--- a/spec/views/organizations/index.html.erb_spec.rb
+++ b/spec/views/organizations/index.html.erb_spec.rb
@@ -3,12 +3,47 @@
 require 'rails_helper'
 
 RSpec.describe 'organizations/index', type: :view do
+  let(:org1) { create(:organization, name: 'Organization 1') }
+  let(:org2) { create(:organization, name: 'Organization 2') }
+  let(:org3) { create(:organization, name: 'Organization 3', provider: false) }
+
   before do
-    assign(:organizations, [create(:organization, name: 'Name1'), create(:organization)])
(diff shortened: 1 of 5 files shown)


spec/views/organizations/index.html.erb_spec.rb
```diff
@@ -3,12 +3,47 @@
...
+    assign(:organizations, [org1, org2, org3])
   end
 
   it 'renders a list of organizations' do
     render
-    assert_select 'tr>td', text: 'Name1'.to_s
+
+    assert_select 'tr>td', text: 'Organization 1'
+    assert_select 'tr>td', text: 'Organization 2'
+    assert_select 'tr>td', text: 'Organization 3'
```
marlo-longley, line 19: We know providers shouldn't show up in this list and there is a ticket to address it; the test should change with that PR? 
I just wanted to verify the current behavior, which also lined up with the doc.
Can remove if it's not good to add.

marlo-longley: PR #735 fixes this issue; note to change the test to reflect that once merged.

corylown: I think it's okay to add this. I'll fix the test before #735 is merged.

corylown (approved):

PR #744, opened 2022-04-29
OAI-PMH Identify response

thatbudakguy:
- Move main contact email to settings
- Implement OAI-PMH Identify response

### DIFF
--- a/app/controllers/oai_controller.rb
+++ b/app/controllers/oai_controller.rb
@@ -11,6 +11,8 @@ class OaiController < ApplicationController
       render_list_records
     when 'ListSets'
       render_list_sets
+    when 'Identify'
+      render_identify
     else
       render_error('badVerb')
     end
@@ -38,6 +40,14 @@ class OaiController < ApplicationController
     headers['X-Accel-Buffering'] = 'no'
 
     render xml: build_list_sets_response(Organization.providers)
+  end
+
+  def render_identify
+    headers['Cache-Control'] = 'no-cache'
+    headers['Last-Modified'] = Time.current.httpdate
+    headers['X-Accel-Buffering'] = 'no'
+
+    render xml: build_identify_response(Upload.order(:created_at).first.created_at)
   end
 
   def render_error(code)
@@ -86,7 +96,9 @@ class OaiController < ApplicationController
   end
   # rubocop:enable Metrics/AbcSize
 
+  # rubocop:disable Metrics/AbcSize
   # rubocop:disable Metrics/MethodLength
+  # See https://www.openarchives.org/OAI/openarchivesprotocol.html#ListRecords
   def build_list_records_response(dump, next_dump = nil)
     Enumerator.new do |yielder|
       yielder << <<~EOXML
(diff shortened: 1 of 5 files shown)


app/controllers/oai_controller.rb
```diff
@@ -40,6 +42,14 @@ def render_list_sets
     render xml: build_list_sets_response(Organization.providers)
   end
 
+  def render_identify
+    headers['Cache-Control'] = 'no-cache'
+    headers['Last-Modified'] = Time.current.httpdate
+    headers['X-Accel-Buffering'] = 'no'
+
+    render xml: build_identify_response(Upload.order(:created_at).first.created_at)
```
thatbudakguy, line 50: not sure this is right, but my understanding isn't very deep...we couldn't have a `created_at` that comes _after_ an `updated_at` though, so it does seem like this should give the earliest known upload from a provider.

app/controllers/oai_controller.rb
```diff
@@ -134,6 +144,27 @@ def build_list_sets_response(organizations)
...
   end
+
+  # See https://www.openarchives.org/OAI/openarchivesprotocol.html#Identify
+  def build_identify_response(earliest_date)
+    Nokogiri::XML::Builder.new do |xml|
+      xml.send :'OAI-PMH', oai_xmlns do
+        xml.responseDate Time.zone.now.iso8601
+        xml.request(oai_params.to_hash) do
+          xml.text oai_url
+        end
```
thatbudakguy, line 155: i have an idea about how to refactor this out now that it's repeated in a few responses; will implement in next PR.

app/controllers/oai_controller.rb
corylown: I'm not really sure either. This gets at our data retention policy and I'm not sure we're clear on that yet. Since we're building the oai response from the normalized dump from the default stream (for the moment, anyway) I wonder if we should use that to get the earliestDateTimestamp.
```suggestion
    render xml: build_identify_response(Stream.default.flat_map { |stream| stream.normalized_dumps }.reject { |dump| dump.oai_xml.nil? }.sort_by(&:created_at).first.created_at)
```

thatbudakguy: Makes sense to me. I'm gonna see if I can do something like this on the ActiveRecord side, but if I can't make a reasonable query I'll use this method.

app/controllers/oai_controller.rb
corylown: Maybe something like this?

```
  def oai_template(xml)
      xml.send :'OAI-PMH', oai_xmlns do
        xml.responseDate Time.zone.now.iso8601
        xml.request(oai_params.to_hash) do
          xml.text oai_url
        end
        yield(xml)
      end
  end

  # See https://www.openarchives.org/OAI/openarchivesprotocol.html#Identify
  def build_identify_response(earliest_date)
    Nokogiri::XML::Builder.new do |xml|
     oai_template(xml) do
        xml.Identify do
          xml.repositoryName t('layouts.application.title')
          xml.baseURL oai_url
          xml.protocolVersion '2.0'
          xml.earliestDatestamp earliest_date.strftime('%F')
          xml.deletedRecord 'transient'
          xml.granularity 'YYYY-MM-DD'
          xml.adminEmail Settings.contact_email
        end
      end
    end.to_xml
  end
  # rubocop:enable Metrics/AbcSize
  # rubocop:enable Metrics/MethodLength
```

corylown: I didn't read carefully enough...thought you wrote "have _no_ idea"

thatbudakguy: This was the idea :)

app/controllers/oai_controller.rb
thatbudakguy: Updated with a query that I think is right; I think the join to `oai_xml_attachment` gets the `.reject()` "for free"

corylown (approved):

Issue #719, opened 2022-04-27
Only allow POD Admins to destroy/remove an organization.


marlo-longley:
Relates to #634 , conflicts with spreadsheet which specifies that only POD Admin should have destroy option.
Noted in spreadsheet

![Screen Shot 2022-04-27 at 15 45 18](https://user-images.githubusercontent.com/1328900/165643569-70b7b6a4-d2c3-43b5-a1cb-531beb604702.png)

ggeisler:
I vote for this being a POD admin only power.

I think these buttons are just remnants of the original table when we didn't have the extra layer of privileges. I would also suggest we remove Edit from this table, even though the Org owner does have edit privileges. I think it creates confusion to have the same action appear in too many places, and the org owner can easily edit their organization via their org-specific homepage links or by simply selecting their provider name from the list in that table.

corylown:
I agree that only POD Admins should be able to destroy an organization.

marlo-longley:
Closed by #748

PR #748, opened 2022-04-29
Align destroy capabilities

### LINKED ISSUE #738: Destroy stream button not present for org owners
Relates to #634 

The spreadsheet + our current ability model indicates that org owners can destroy non-default streams of their own org (Owners have the `:crud` ability on Stream). While the intended Destroy button is currently showing up for POD Admins, the button isn't showing up for org owners.

Current button code:
https://github.com/pod4lib/aggregator/blob/f42b15f90c770258cf8b9963d55de8c32ec7955d/app/views/streams/index.html.erb#L31

If you change `if can? :delete, stream` to `if can? :destroy, stream`, the button properly shows up for owners.
I don't know the underlying implications of `:delete` vs `:destroy` 

This ticket should also include looking through the UI to other places where the destroy/delete button is supposed to be showing, and probably writing a test.

marlo-longley:
Fixes #738 

Also found a decent way to test the providers scope on organizations.

### DIFF
--- a/spec/models/organization_spec.rb
+++ b/spec/models/organization_spec.rb
@@ -48,4 +48,16 @@ RSpec.describe Organization do
       expect(organization.provider?).to be true
     end
   end
+
+  describe '#providers' do
+    it 'includes providers in scope' do
+      org = described_class.create!(provider: true)
+      expect(described_class.providers).to include(org)
+    end
+
+    it 'excludes non-providers in scope' do
+      org = described_class.create!(provider: false)
+      expect(described_class.providers).not_to include(org)
+    end
+  end
 end
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -49,25 +49,29 @@ class Ability
     end
 
     can :manage, :all if user.has_role?(:admin)
-    cannot :delete, Stream, default: true
     can :read, Organization, public: true
     can :manage, :dashboard_controller if user.has_role?(:admin)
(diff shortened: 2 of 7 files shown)


marlo-longley:
Latest commit also fixes #719

spec/models/organization_spec.rb
```diff
@@ -48,4 +48,16 @@
...
+    it 'includes providers in scope' do
+      org = described_class.create!(provider: true)
+      expect(described_class.providers).to include(org)
+    end
+
+    it 'excludes non-providers in scope' do
+      org = described_class.create!(provider: false)
+      expect(described_class.providers).not_to include(org)
+    end
+  end
```
thatbudakguy, line 62: this is sort of what i was alluding to in standup — if you define a super simple scope using ActiveRecord on one field, there isn't really a way for it to fail unless Rails's own code fails, so you're effectively testing Rails. not a bad thing necessarily, just generally less useful than other kinds of tests

thatbudakguy (approved):
pull request title sounds like a metal band. code looks good 👍

corylown (approved):

spec/models/organization_spec.rb
marlo-longley: @thatbudakguy fair! I couldn't think of another way to test it.

marlo-longley: also - destroy permissions!!!

PR #749, opened 2022-04-29
Send confirm email only when email field is updated; fix flash message.

### LINKED ISSUE #722: Updating organization contact field causes multiple confirmation emails to be sent
This was reported by Duke. They changed their organization contact email and received 7 confirmation emails within a few minutes. I confirmed this behavior, although I only received two emails.

### LINKED ISSUE #723: After updating an organization's contact link and clicking on the confirm link in the email there is no message in the UI to indicate success.
The user is just taken to the organization page with no indication of whether the confirmation succeeded or failed. A flash message with user feedback would be helpful.

corylown:
I'm ignoring the issue of whether we care if the contact email gets confirmed or not before displaying it. Right now it displays right away. Fixes #722 and Fixes #723

### DIFF
--- a/spec/models/contact_email_spec.rb
+++ b/spec/models/contact_email_spec.rb
@@ -21,11 +21,14 @@ RSpec.describe ContactEmail, type: :model do
     end
   end
 
+  # rubocop:disable RSpec/SubjectStub
   it 'sends a confirmation email' do
+    allow(contact_email).to receive(:saved_change_to_attribute?).with(:email).and_return(true)
     expect do
-      contact_email.save
+      contact_email.update(email: 'someone@example.com')
     end.to have_enqueued_mail(ContactEmailsMailer, :confirm_email)
   end
+  # rubocop:enable RSpec/SubjectStub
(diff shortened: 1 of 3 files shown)


thatbudakguy (approved):
displaying an unconfirmed email seems like an OK outcome for now. thx for fixing these

spec/models/contact_email_spec.rb
```diff
@@ -21,14 +21,19 @@
     end
   end
 
+  # rubocop:disable RSpec/SubjectStub
   it 'sends a confirmation email' do
+    allow(contact_email).to receive(:saved_change_to_attribute?).with(:email).and_return(true)
```
corylown, line 26: This is probably a bad idea, but I cannot get the `saved_change_to_attribute?(:email)` to return true just by updating the `contact_email.email`. Anyone have ideas?

thatbudakguy: decided in standup yesterday that we're fine with this, i think

PR #757, opened 2022-05-02
Remove container wrapper and add margin to main menu

### LINKED ISSUE #721: Topbar spacing is off when using Safari
If I remember correctly, we already did some topbar element spacing fixes, but I still some issues using Safari (MacOS) on -prod. Unless this has been fixed but not deployed, here's the extra horizontal space that needs reducing:

- [ ] Between the "POD" and the "Aggregator" part of the POD logo
- [ ] Between the POD logo and the main menu items

Screen Shot 2022-04-28 at 9 25 34 AM

ktamaral:
Fixes #721

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -170,3 +170,8 @@ input[type=file][data-direct-upload-url][disabled] {
   background-color: $warning;
   max-width: 100vw;
 }
+
+nav.navbar {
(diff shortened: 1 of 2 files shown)


ggeisler:
Potential alternative solution (in case my comment on that ticket isn't very noticeable): https://github.com/pod4lib/aggregator/issues/721#issuecomment-1115416460

Issue #711, opened 2022-04-27
OAI: respond to ListRecords requests 1000 records at a time


corylown:
- Respond in chunks (~1000 records at a time, with resumption tokens)
- Chunks could be filtered portions of pre-generated file, or a series of pre-generated files in a folder/directory

thatbudakguy:
do we support `ListRecords` requests that don't specify a set (i.e. target all records in the aggregator)?

thatbudakguy:
just saw john's comment on #569:
> We're aiming to have one set per provider organization (not per stream). OAI clients can request harvesting in full or with time-bounds from any of those sets. They can also make a request that doesn't specify a set.

so, `ListRecords` with no set should be supported.

JohnMarkOckerbloom:
Just a note on this ticket: While I do think we need to support set-less ListRecords requests, I'm not aware of Index Data needing them, so it's higher priority to get institutional-set ListRecords requests working than setless ones.

PR #761, opened 2022-05-03
OAI-PMH refactor & ListMetadataFormats

thatbudakguy:
- Ensure earlestDatestamp displays with no items
- Refactor OaiController and extract OaiConcern
- Implement OAI-PMH ListMetadataFormats response

### DIFF
--- a/app/controllers/oai_controller.rb
+++ b/app/controllers/oai_controller.rb
@@ -3,19 +3,26 @@
 # Produce OAI-PMH responses
 # rubocop:disable Metrics/ClassLength
 class OaiController < ApplicationController
-  load_and_authorize_resource :organization
+  include OaiConcern
+
+  skip_authorization_check
+  rescue_from OaiConcern::OaiError, with: :render_error
 
   def show
-    case params[:verb]
-    when 'ListRecords'
-      render_list_records
-    when 'ListSets'
-      render_list_sets
-    when 'Identify'
-      render_identify
-    else
-      render_error('badVerb')
-    end
+    verb = params.require(:verb)
+    send("render_#{verb.underscore}")
+  rescue ActionController::ParameterMissing
+    raise OaiConcern::BadVerb
+  end
+
+  def method_missing(method, *_args, &_block)
+    raise OaiConcern::BadVerb if method.to_s.start_with?('render_')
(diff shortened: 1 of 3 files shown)


thatbudakguy:
this doesn't address #711 but I figured we should either merge or change it before I start on that work.

app/controllers/oai_controller.rb
```diff
@@ -170,8 +184,32 @@ def build_identify_response(earliest_date)
...
+
+  # See http://www.openarchives.org/OAI/openarchivesprotocol.html#ListMetadataFormats
+  def build_list_metadata_formats_response
+    Nokogiri::XML::Builder.new do |xml|
+      build_oai_response xml, list_metadata_formats_params do
+        xml.ListMetadataFormats do
+          xml.metadataFormat do
+            xml.metadataPrefix 'marc21'
+            xml.schema 'http://www.loc.gov/standards/marcxml/schema/MARC21slim.xsd'
+            xml.metadataNamespace 'http://www.loc.gov/MARC21/slim'
```
thatbudakguy, line 196: i can't understand the difference between marc21 and marcxml, or whether this is right...

app/controllers/oai_controller.rb
```diff
@@ -92,16 +100,29 @@ def list_record_normalized_dump_candidates
...
+      xml.request(params.to_hash) do
+        xml.text oai_url
+      end
+      yield xml
+    end
+  end
+
   # rubocop:disable Metrics/MethodLength
   # See https://www.openarchives.org/OAI/openarchivesprotocol.html#ListRecords
   def build_list_records_response(dump, next_dump = nil)
```
corylown, line 128: The `build_request` method is now gone, but it's being called from line 137 in the `build_oai_response` method.

app/controllers/oai_controller.rb
corylown: I think what you have is correct. The terminology is confusing, but MARC21slim is an XML expression of MARC21.

corylown (changes requested):
There's a method being called that's now missing -- see inline comments. Otherwise this looks good to me.

corylown (approved):

PR #762, opened 2022-05-03
Use single layout for streams and Provider home

### LINKED ISSUE #755: All streams should use new provider page layout
I guess I didn't explicitly say this in my tickets for redesigning the provider page to use tabs for Uploaded files, Normalized data, Processing status, etc. but that layout applies to any stream, not just the default stream. Currently we use this layout for the default stream:

Screen Shot 2022-05-02 at 10 49 17 AM

---

But if you use the View/Manage streams link and select another stream from the View/Manage streams page, you see this:

Screen Shot 2022-05-02 at 10 54 21 AM

---

where the uploaded files table is below the processing status section, and the Normalized data and MARC analysis sections are not shown at all (maybe the Normalized data section doesn't apply to non-default streams? but the MARC analysis at least does, I think?).

The idea with the provider page redesign was that any stream uses the new layout with the tabbed pages and file stats box above the tabs (consistent presentation for the user), rather than using the old way of displaying all sections on a single page. 

I'd suggest we omit the "Make default" button on the non-default stream version of the page and require the user to make that change from the Manage streams page.

marlo-longley:
Fixes #755 

The one odd thing about this solution, which may not be a problem is:
1. navigate to Provider home (`/organizations/stanford`) which defaults of the first Uploaded files tab
2. click on another tab like Processing status
3. Now click back to your original tab, Uploaded files
--- you can see that you are now on the stream page version of Uploaded files (`/organizations/stanford/streams/234`) rather than the Provider home version at `/organizations/stanford`

Is this a problem? There is still a place to land at `/organizations/stanford`, but once you start clicking around, you are within the /streams URL zone. I wasn't coming up with a way to avoid this other than duplicating the routes within both organization and streams controllers, which seemed silly.


app/models/ability.rb
```diff
@@ -43,9 +43,10 @@ def initialize(user, token = nil)
     if user.roles.any?
       can :read, ActiveStorage::Attachment, { record: { organization: { public: true } } }
       can :read, MarcRecord, upload: { organization: { public: true } }
-      can %i[read profile info], [Stream, Upload], organization: { public: true }
+      can %i[read profile normalized_data processing_status], [Stream], organization: { public: true }
+      can %i[read info], [Upload], organization: { public: true }
```
corylown, line 47:
```suggestion
      can %i[read profile normalized_data processing_status], Stream, organization: { public: true }
      can %i[read info], Upload, organization: { public: true }
```

app/models/ability.rb
```diff
@@ -56,16 +57,14 @@ def initialize(user, token = nil)
     owned_orgs = Organization.with_role(:owner, user).pluck(:id)
     can :manage, Organization, id: owned_orgs
     cannot :destroy, Organization, id: owned_orgs
-    can %i[crud profile], [Stream], organization: { id: owned_orgs }
-    can %i[crud info], [Upload], organization: { id: owned_orgs }
+    can %i[crud], [Stream], organization: { id: owned_orgs }
+    can %i[crud], [Upload], organization: { id: owned_orgs }
```
corylown, line 61:
```suggestion
    can :crud, Stream, organization: { id: owned_orgs }
    can :crud, Upload, organization: { id: owned_orgs }
```

corylown (commented):
A couple of minor inline suggestions. The routing switching between organization#show and various stream routes does seem a little odd, but I'm not sure right now how to resolve that. I think we could go with this for now and see if a different solution to the routing emerges.

corylown (approved):

PR #775, opened 2022-05-05
Redesign uploads index page

### LINKED ISSUE #745: Update content and layout of Uploads index page
Related to #746.

The current [Uploads index page](https://pod.stanford.edu/organizations/stanford/uploads/) was just a very basic MVP implementation, I believe, and doesn't fit in well with the current application design. As an example, here's the current version:

Screen Shot 2022-04-29 at 1 05 02 PM

---

Here's a suggested revision:

Screen Shot 2022-04-29 at 12 53 07 PM

---

- [x] **A:** Change the current heading "Uploads" to an `<h2>` (it is currently an `<h1`>).
- [x] **B:** Add a `New upload` primary button to replace the "New Upload" link currently at the very bottom of the uploads list. (Note that I'm suggesting `New upload` for the button label, which doesn't match the mockup; I changed my mind and forgot to update the mockup.)

Update the current table as described below:
- [x] **C:** Name column. Link the name of the upload. This is the link currently used as the "Show" link in the last column of the current table.
- [x] **D:** Add a date column.
- [x] Remove the Files column. The user can see the individual files that are part of an upload on the upload show page.
- [x] **E:** Add a `Stream` column. This isn't in the current table but I assume we know which stream an upload was created in? If so, add the column, displaying the stream name and linking that to the stream page. 
- [x] **F:** Add a column for `Actions` but only display this column to org owners and POD admins (that's my suggestion but permissions for these actions are not yet defined in the spreadsheet).
- [x] _Show_ action: Since the upload name will be linked to its show page, I don't think we should also display the "Show" action that is currently displayed. Remove it from this column.
- [x] **G:** _Edit_ action: Display this as a secondary button, **if** we can verify that this is an actual action that users might need to use. Currently selecting that action for an upload returns an error page.
- [x] **H:** _Delete_ action: If this is something we want a user to be able to do, make this a danger button.
- [x] Sort the table by Date (most recent first). I'm not sure how the current list is sorted but given the upload names are not easy to scan, I think finding an upload by date is likely the best way to organize the rows.
- [x] Use the `table-striped` class as on other tables.
- [ ] We probably should add pagination to this page, similar to how we added it to the Uploaded files table.

marlo-longley:
Fixes #745 

:)

Screen Shot 2022-05-05 at 11 20 47

### DIFF
--- a/app/views/uploads/index.html.erb
+++ b/app/views/uploads/index.html.erb
@@ -4,10 +4,6 @@
 
     <h2 class="font-weight-normal mb-0">Uploads</h2>
 
-    <div class="ms-auto p-2">
-      <span class="ms-4"><%= link_to 'New upload', new_organization_upload_path(@stream.organization, stream: (@stream.default? ? nil : @stream.friendly_id)), class: 'btn btn-primary' if can? :create, @stream.uploads.build %></span>
-    </div>
-
   </div>
 
   <table class="table table-striped mt-5">
@@ -26,7 +22,7 @@
         <tr>
           <td><%= link_to upload.name, [@organization, upload] %></td>
           <td><%= local_time(upload.created_at, format: datetime_display_format()) %></td>
-          <td><%= link_to @stream.display_name, organization_stream_path(@stream.organization, @stream) %></td>
+          <td><%= link_to upload.stream.display_name, organization_stream_path(upload.stream.organization, upload.stream) %></td>
           <td><%= link_to 'Destroy', [@organization, upload], method: :delete, data: { confirm: 'Are you sure?' }, class: 'btn btn-danger btn-sm' if can? :destroy, upload %></td>
         </tr>
       <% end %>
--- a/app/views/uploads/index.html.erb
+++ b/app/views/uploads/index.html.erb
@@ -1,31 +1,37 @@
-<%= render 'shared/organization_header' %>
+<%= render 'shared/layout_upload_page' do %>
 
-<div class="container">
-  <h1>Uploads</h1>
+  <div class="d-flex align-items-center">
 
-  <table class="mt-4">
+    <h2 class="font-weight-normal mb-0">Uploads</h2>
+
+    <div class="ms-auto p-2">
+      <span class="ms-4"><%= link_to 'New upload', new_organization_upload_path(@stream.organization, stream: (@stream.default? ? nil : @stream.friendly_id)), class: 'btn btn-primary' if can? :create, @stream.uploads.build %></span>
+    </div>
+
+  </div>
+
+  <table class="table table-striped mt-5">
+
     <thead>
       <tr>
         <th>Name</th>
-        <th>Files</th>
-        <th colspan="3"></th>
+        <th>Date</th>
+        <th>Stream</th>
+        <th>Actions</th>
       </tr>
     </thead>
 
     <tbody>
-      <% @uploads.each do |upload| %>
+      <% @uploads.sort_by(&:created_at).reverse.each do |upload| %>
         <tr>
-          <td><%= upload.name %></td>
-          <td><%= upload.files %></td>
-          <td><%= link_to 'Show', [@organization, upload] %></td>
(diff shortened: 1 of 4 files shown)


app/views/uploads/index.html.erb
```diff
@@ -1,31 +1,37 @@
...
+      <% @uploads.sort_by(&:created_at).reverse.each do |upload| %>
         <tr>
-          <td><%= upload.name %></td>
-          <td><%= upload.files %></td>
-          <td><%= link_to 'Show', [@organization, upload] %></td>
-          <td><%= link_to 'Edit', edit_organization_upload_path(@organization, upload) %></td>
-          <td><%= link_to 'Destroy', [@organization, upload], method: :delete, data: { confirm: 'Are you sure?' } %></td>
+          <td><%= link_to upload.name, [@organization, upload] %></td>
+          <td><%= local_time(upload.created_at, format: datetime_display_format()) %></td>
+          <td><%= link_to @stream.display_name, organization_stream_path(@stream.organization, @stream) %></td>
```
corylown, line 29: I don't think this is quite right. We need the stream associated with the upload -- maybe something like:

```suggestion
          <td><%= link_to upload.stream.display_name, organization_stream_path(upload.stream.organization, upload.stream) %></td>
```

app/views/uploads/index.html.erb
```diff
@@ -1,31 +1,37 @@
...
+<%= render 'shared/layout_upload_page' do %>
 
-<div class="container">
-  <h1>Uploads</h1>
+  <div class="d-flex align-items-center">
+
+    <h2 class="font-weight-normal mb-0">Uploads</h2>
+
+    <div class="ms-auto p-2">
+      <span class="ms-4"><%= link_to 'New upload', new_organization_upload_path(@stream.organization, stream: (@stream.default? ? nil : @stream.friendly_id)), class: 'btn btn-primary' if can? :create, @stream.uploads.build %></span>
```
corylown, line 8: I'm a little concerned about providing an upload button here. How do we know which stream the upload will be associated with?

corylown (changes requested):
See inline suggestions/questions.

app/views/uploads/index.html.erb
corylown: There's a similar issue on the upload show page. I'd be okay merging this as-is and handling the ambiguity of the "New upload" button as a separate design issue.

marlo-longley: OK. This was in the mockup so I added the button.
I reused some existing code that determines the stream here:
` stream: (@stream.default? ? nil : @stream.friendly_id))`

This New upload button code is the same all over the app, including on the Provider home page. The route `http://localhost:3000/organizations/stanford/uploads/new` is the upload page, and has no indication of what stream you're in.

app/views/uploads/index.html.erb
marlo-longley: Ahh this was leftover from the old code! Thanks.

marlo-longley: @corylown actually I take this back - I added the @stream variable in the controller. It should be accurate as is. ?

app/views/uploads/index.html.erb
corylown: I created a new issue for this so we don't have to worry about for this PR: https://github.com/pod4lib/aggregator/issues/781

app/views/uploads/index.html.erb
corylown: I don't think the controller variable works in this context because the controller only grabs one stream (haven't looked closely at the logic, but in the table you're showing uploads from all different streams so you need to show the stream associated with each separate upload.

marlo-longley: Ok makes sense thank you.

marlo-longley: @corylown actually you are right below; fixing this stream issue now breaks the New upload button. I am going to remove it from the layout.

corylown (approved):

PR #780, opened 2022-05-05
The "Part of stream" link should use the stream associated with the current upload

### LINKED ISSUE #779: Upload index page link to "Part of #{stream}" always links to default stream even if upload was to a different stream.

corylown:
Fixes #779


marlo-longley:
Thanks for catching that!

PR #783, opened 2022-05-06
Refactor new upload buttons on upload index & show pages

### LINKED ISSUE #781: "New upload" buttons on upload index and show pages are ambiguous
@ggeisler I noticed this while reviewing PRs for #770 & #775. The "New upload" buttons/links are ambiguous.

## Upload Index

On the index page the uploads shown can be associated with any number of different streams. It's unclear in this context which stream a new upload would be associated with. The current behavior is that the upload will be associated with the current default stream. I think we should indicate this somehow in the UI. The solution might be as simple as changing the text on the button to: "New upload to default stream" -- but maybe that's too long.
Screen Shot 2022-05-05 at 5 44 09 PM

## Upload Show

On the show page it's especially confusing when you're viewing an upload from a stream that's not the default because you might be expecting to add an upload to the stream associated with the upload you're currently viewing. But that's not what happens. Just like on the index page it adds the upload to the current default stream. I think I'd suggest either removing the "New upload" button from the show page or clarify the button text here as well.
Screen Shot 2022-05-05 at 5 43 46 PM

corylown:
Closes #781

### DIFF
--- a/app/controllers/uploads_controller.rb
+++ b/app/controllers/uploads_controller.rb
@@ -14,9 +14,7 @@ class UploadsController < ApplicationController
 
   # GET /uploads/1
   # GET /uploads/1.json
-  def show
-    @stream = current_stream
-  end
+  def show; end
 
   # GET /uploads/new
   def new; end
--- a/app/views/uploads/index.html.erb
+++ b/app/views/uploads/index.html.erb
@@ -4,6 +4,10 @@
 
     <h2 class="font-weight-normal mb-0">Uploads</h2>
 
+    <div class="ms-auto p-2">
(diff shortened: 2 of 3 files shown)


marlo-longley:
@corylown thanks for fixing this, I have 2 questions -
1. Is it intended that there is no "New upload" button on non-default uploads#show pages?
Screen Shot 2022-05-06 at 09 09 03


2. When you click New upload to default stream from the uploads#index page and land on the upload page, I think it could be helpful in a future PR to indicate you are on the default stream - it's ambigous on the page
Screen Shot 2022-05-06 at 09 09 37 -- maybe a new issue would be good depending on what you think

marlo-longley (approved):

PR #799, opened 2022-05-10
Refactor DefaultStreamHistories

### LINKED ISSUE #789: Fix foreign key relationships for default stream history.
Deleting an organization causes a foreign key violation. I think the foreign key relationship between organization and default stream history is an error and the relationship should be through streams instead.

corylown:
The relationships between DefaultStreamHistory, Stream, and Organization were not quite right. This adjusts the model relationships to prevent foreign key violations (when deleting an organization, etc.). It also moves to the Stream model the method for getting the previous default stream history object for a stream. Adds test coverage.
Fixes #789


marlo-longley (approved):
@corylown thanks for your help :)

PR #800, opened 2022-05-10
Refactor InterstreamDelta model relationships

### LINKED ISSUE #796: Fix InterstreamDelta model
Should belong to normalized dump not stream.

corylown:
Fixes #796


app/jobs/generate_interstream_delta_job.rb
```diff
@@ -100,7 +100,7 @@ def perform(stream)
 
     # Attach Files
     unless current_stream_dump.interstream_delta
-      current_stream_dump.interstream_delta = InterstreamDelta.create(normalized_dump: current_stream_dump, stream: stream)
+      current_stream_dump.interstream_delta = InterstreamDelta.create(normalized_dump: current_stream_dump)
```
cbeer, line 103: Can't ActiveRecord do this for us?

```suggestion
  current_stream_dump.create_interstream_delta
```

(untested.)

corylown: Seems to work. Thanks!

PR #801, opened 2022-05-11
Implement badges with stream history dates

### LINKED ISSUE #786: Update stream/provider show page with status badge and tooltip
This is a more actionable version of #771, since we've made a decision about which option to go with.

When the user is viewing the stream show page for a non-default stream, we want the user to know if that stream was ever the default stream, and if so, give the user a way to know the date range(s) it was the default stream. There's also a visual layout improvement we can make to the stream metadata box when working on this page.

## Move the "Last updated" info into the stream metadata box
This will make the page header info a little cleaner and enable us to make the Stream, Upload, and File show pages a bit more consistent in the header/metadata area.

- [x] Move the "Last updated:" line from below the page header to the top of the stream metadata box, centered
- [x] Add a divider or bottom border to separate that line from the file/records metadata elements

either-option

---

## Update the stream status badge 
Currently, when a stream is the default stream, we indicate that status with a "Default" badge next to the stream name in the page heading. We don't do anything similar for streams that were previously the default stream, but have decided to do that. (If a stream was never the default we won't show a badge next to its name.)

Screen Shot 2022-05-06 at 10 47 51 AM

---

Screen Shot 2022-05-06 at 10 48 26 AM

- [x] When displaying the stream show page for a stream that was previously the default stream, display a badge next to the stream name with the label `Previous default`. Use the `badge bg-warning text-dark` to style the badge.
- [x] Add the `text-dark` class to the `bg-info` badge used for the default stream, so the label text is dark rather than white.
- [x] For both badge cases, add an `info-circle-fill` icon to the badge.

## Add a tooltip to the stream status badges
With an info icon in the badge, we have a place to add an affordance for the user who might want to know when a stream was previously the default stream, or when the current default stream became the default stream. In #771 we decided to go with the [Bootstrap tooltip option](https://getbootstrap.com/docs/5.0/components/tooltips/) for this.

Screen Shot 2022-05-06 at 10 54 05 AM

---

Screen Shot 2022-05-06 at 10 54 32 AM

---

Screen Shot 2022-05-06 at 10 55 17 AM

---

- [x] Trigger a Bootstrap tooltip by mouseover/keyboard selection of the info icon in the stream status badge.
- [x] If the stream is a previous default, display the start and end date/time values separated by "`to`.
- [x] If the stream is the current default, use the text `Default since` followed by the date/time it became the default stream.
- [x] A previous default stream might have more than one date range when it was previously the default. If so, display each date range on a separate line.

Note that the requirement to display multiple date ranges means we will likely need to use HTML elements inside the tooltip markup, such as wrapping each date range line in a `<div>` or `<p>`.  Also, to ensure the tooltips work for keyboard and screenreaders, we might have to wrap the badge tooltip in a link. See https://getbootstrap.com/docs/5.0/components/tooltips/#markup for more about this.

### LINKED ISSUE #787: Update Streams index page table, add tooltip for status badges
Related to #786, in that we want to use the same badge/tooltip approach as described in that ticket. Aside that that part, there are a handful of minor table arrangement updates.

## Table updates

Screen Shot 2022-05-06 at 11 20 02 AM

- [x] **A:** Move the "Name" column to be the first two column of the table. 
- [x] **B:** Add a `Status` heading for the second column (it doesn't currently have a heading, and is the first column).
- [x] **C:** Add an `Actions` column heading, centered, for the last column that displays the "Make default" and "Destroy" buttons (if the user has permissions to those).
- [x] **C:** Swap order of the the "Make default" and "Destroy" buttons.
- [x] **D:** I'm not sure if this is represented in another ticket, but I believe we want to prevent any stream that was previously a default from being made the default again. So we should not display the "Make default" button for previously default streams.
- [x] **E:** Change the label of the "Destroy" button to "Delete".
- [x] **F:** When a stream was previously the default stream, add a `Previous default` badge to its Status column. Styling is `badge bg-warning text-dark` (I forgot to update the badge label color in my mockup, but it should be dark).
- [x] **F:** Update the current default stream badge with the `text-dark` class so its label is also dark text.
- [x] **F:** For both badges, add an `info-circle-fill` icon to the badge.
- [x] I'm not sure how the current table is sorted, but it probably should be changed to sort by the "Last updated" column, most recent first (my mockup doesn't show it that way, just because I was following the current implementation).

## Add tooltip to badge to display default stream date range

Screen Shot 2022-05-06 at 11 32 43 AM

---

Screen Shot 2022-05-06 at 11 32 58 AM

---

This is the same as described in #786:

- [x] Trigger a Bootstrap tooltip by mouseover/keyboard selection of the info icon in the stream status badge.
- [x] If the stream is a previous default, display the start and end date/time values separated by "`to`.
- [x] If the stream is the current default, use the text `Default since` followed by the date/time it became the default stream.
- [x] A previous default stream might have more than one date range when it was previously the default. If so, display each date range on a separate line.

Note that the requirement to display multiple date ranges means we will likely need to use HTML elements inside the tooltip markup, such as wrapping each date range line in a `<div>` or `<p>`.  Also, to ensure the tooltips work for keyboard and screenreaders, we might have to wrap the badge tooltip in a link. See https://getbootstrap.com/docs/5.0/components/tooltips/#markup for more about this.

marlo-longley:
Fixes #786 
Fixes #787

### DIFF
--- a/app/views/streams/_header.html.erb
+++ b/app/views/streams/_header.html.erb
@@ -2,7 +2,10 @@
   <h2 class="font-weight-normal mb-0">Stream: <%= stream.display_name %></h2>
 
   <div class="p-2">
-    <%= content_tag :span, 'Default', class: 'badge bg-info me-3' if stream.default? %>
+    <%= default_stream_status_badge(stream) if stream.default? %>
+    <% if stream.default_stream_histories.any? && stream.default == false %>
+      <%= previous_default_stream_status_badge(stream) %>
+      <% end %>
     <%= link_to can?(:edit, stream) ? t('.manage_streams') : t('.view_streams'), organization_streams_path(stream.organization) %>
     <span class="badge bg-secondary"><%= stream.organization.streams.active.count %></span>
   </div>
@@ -12,11 +15,6 @@
     <%= link_to t('.reanalyze'), reanalyze_organization_stream_path(stream.organization, stream), method: :post, class: 'btn btn-secondary align-self-center' if can?(:manage, stream) %>
   </div>
 </div>
-
-<p class="lead">
-  <% stream_updated = stream.uploads.last&.updated_at || stream.updated_at %>
-  Last updated: <%= content_tag :span, "#{time_ago_in_words(stream_updated)} ago", title: local_time(stream_updated, format: datetime_display_format()) %>
-</p>
 
 <div class="mb-3">
   <% unless stream.default? %>
--- a/app/views/streams/_header.html.erb
+++ b/app/views/streams/_header.html.erb
@@ -5,7 +5,7 @@
     <%= default_stream_status_badge(stream) if stream.default? %>
     <% if stream.default_stream_histories.any? && stream.default == false %>
       <%= previous_default_stream_status_badge(stream) %>
-      <% end %>
+    <% end %>
     <%= link_to can?(:edit, stream) ? t('.manage_streams') : t('.view_streams'), organization_streams_path(stream.organization) %>
     <span class="badge bg-secondary"><%= stream.organization.streams.active.count %></span>
   </div>
--- a/app/views/streams/_header.html.erb
+++ b/app/views/streams/_header.html.erb
@@ -2,10 +2,7 @@
   <h2 class="font-weight-normal mb-0">Stream: <%= stream.display_name %></h2>
 
   <div class="p-2">
-    <%= default_stream_status_badge(stream) if stream.default? %>
-    <% if stream.default_stream_histories.any? && stream.default == false %>
-      <%= previous_default_stream_status_badge(stream) %>
-    <% end %>
+    <%= default_stream_status_badge(stream) if stream.default_stream_histories.any? %>
     <%= link_to can?(:edit, stream) ? t('.manage_streams') : t('.view_streams'), organization_streams_path(stream.organization) %>
     <span class="badge bg-secondary"><%= stream.organization.streams.active.count %></span>
   </div>
--- a/app/assets/stylesheets/bootstrap-overrides.scss
+++ b/app/assets/stylesheets/bootstrap-overrides.scss
@@ -8,3 +8,5 @@
     width: 100%;
   }
 }
+
+$tooltip-max-width: 600px;
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -24,4 +24,39 @@ module ApplicationHelper
   def datetime_display_format
     '%B %e, %Y %l:%M%P %Z'
   end
+
+  def hidden_time(time, time_type)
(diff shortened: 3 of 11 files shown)


app/views/streams/_header.html.erb
```diff
@@ -2,7 +2,10 @@
   <h2 class="font-weight-normal mb-0">Stream: <%= stream.display_name %></h2>
 
   <div class="p-2">
-    <%= content_tag :span, 'Default', class: 'badge bg-info me-3' if stream.default? %>
+    <%= default_stream_status_badge(stream) if stream.default? %>
```
corylown, line 5: This isn't displaying all the start/end times for the currently default stream if it was the default more than once. I have 2 streams with the following histories:

```
[[#<DefaultStreamHistory:0x00007f1453af2878                  
   id: 6,                                                    
   stream_id: 6,                                             
   start_time: Wed, 11 May 2022 19:06:41.991248000 UTC +00:00,
   end_time: Thu, 12 May 2022 20:15:20.589156000 UTC +00:00>,
  #<DefaultStreamHistory:0x00007f1453af2670
   id: 8,
   stream_id: 6,
   start_time: Thu, 12 May 2022 20:16:00.086639000 UTC +00:00,
   end_time: Thu, 12 May 2022 20:19:18.976072000 UTC +00:00>],
 [#<DefaultStreamHistory:0x00007f1453abf1f8
   id: 5,
   stream_id: 5,
   start_time: Wed, 11 May 2022 11:53:16.471164000 UTC +00:00,
   end_time: Wed, 11 May 2022 19:06:41.967913000 UTC +00:00>,
  #<DefaultStreamHistory:0x00007f1453abf018
   id: 7,
   stream_id: 5,
   start_time: Thu, 12 May 2022 20:15:20.620490000 UTC +00:00,
   end_time: Thu, 12 May 2022 20:16:00.068557000 UTC +00:00>,
  #<DefaultStreamHistory:0x00007f1453abee10
   id: 9,
   stream_id: 5,
   start_time: Thu, 12 May 2022 20:19:18.994331000 UTC +00:00,
   end_time: nil>]]
```

I'd expect the current default (stream 5) to show three time ranges, but it only shows the start time for the current default history. The previous history start and end times are not showing. 
Screen Shot 2022-05-12 at 4 23 02 PM


It is correctly showing the start and end times for stream 6.
Screen Shot 2022-05-12 at 4 23 07 PM

corylown (changes requested):

marlo-longley:
@corylown thank you for reviewing! 
I guess the reason for that is the following UI instruction from Gary:
`If the stream is the current default, use the text Default since followed by the date/time it became the default stream.`
So I didn't even consider adding those previous histories. Since I'm doing it already for the previous default/yellow badge, i should be able to fix this pretty easily. Thanks!

marlo-longley:
Screen Shot 2022-05-12 at 15 47 24

corylown (approved):

PR #808, opened 2022-05-12
Update design of uploads info page

### LINKED ISSUE #792: Update layout and content of File show page
This is the page you get to either by going to a upload show page and selecting the info magnifying glass icon for one of its listed files, or by going to a stream page and selecting the info magnifying glass icon for one of its listed files (e.g., `uploads/99999/info/55555`)

The current layout of the top part of the page is still a bit MVP-ish and doesn't match some of our recent updates for pages with similar information (e.g., stream show page): 

Screen Shot 2022-05-06 at 2 07 15 PM

---

That's the part I'm suggesting we update here, nothing having to do with the MARC data section below the Summary heading.

Screen Shot 2022-05-06 at 2 07 57 PM

---

- [x] **A:** Add `File:` before the filename in the page heading. This will follow the pattern we're using for the Stream page and make it easier for the user to remember the type of entity they're looking at.
- [x] **B:** Make the "Download" button a `btn-primary` since it's the only real action the user can do on the page (at least via a button). I think sometimes there is also a "Source" button here but #784 indicates there should be one or the other, so whichever is displayed should be `btn-primary`.
- [x] **C:** Add a metadata info box similar to the one on the Stream show page. Ideally these should be styled very similarly, so maybe start by copying the code for that one... I'll can do some fine-tuning of the styling later on, though, since I think there are some details still to work out.
- [x] **D:** Add the file created date/time as a `<p class="lead">` element centered at the top of the box. If #786 hasn't been completed before this ticket is worked on, note that I'm suggesting a similar change there for the Stream show page. We don't show the file's created date/time on this page now, but we should have that data available to add to the page.
- [x] **E:** Move the Size and Records info from below the page title area into the metadata box and change the styling to match the styling of similar data in the Stream show page.
- [x] **F:** There's currently a "Stream" link above the page heading. Move that into the metadata box, using `Part of stream:` as a label preceding it, and displaying the name of the stream as the link label.
- [x] **G:** Similarly, display a link to the upload associated with this file, using `Part of upload:` as a label preceding it. We don't currently display any info about the file's parent upload, but I'm assuming that data is accessible for us to display. I think it might be useful since at least an admin user can arrive at this page by starting from an upload link via the Activity page (but if I understand our current permissions decision, any user should be able to view the upload show page, so there's no reason to restrict this link).

marlo-longley:
Fixes #792

### DIFF
--- a/app/views/uploads/_summary_card.html.erb
+++ b/app/views/uploads/_summary_card.html.erb
@@ -1,12 +1,12 @@
 <div class="card col-12 col-md-8 offset-md-2 col-lg-6 offset-lg-3 col-xxl-4 offset-xxl-4 mb-3">
     <div class="card-body bg-light">
         <div class="row text-center">
-            <p class="lead mb-0">
+            <p class="lead mb-1">
                 Created: <%= local_time(@upload.created_at, format: datetime_display_format) %>
             </p>
             <span class="border-bottom mt-1 mb-3"></span>
         </div>
-        <div class="row">
+        <div class="row mt-3">
             <div class="col">
                 <p>Size: <span class="fw-bold"><%= number_to_human_size @blob.byte_size %></p> 
             </div>
--- a/app/views/uploads/_summary_card.html.erb
+++ b/app/views/uploads/_summary_card.html.erb
@@ -1,4 +1,4 @@
-<div class="card col-12 col-md-8 offset-md-2 col-lg-6 offset-lg-3 col-xxl-4 offset-xxl-4 mb-3">
+<div class="card col-12 col-md-8 offset-md-2 col-lg-6 offset-lg-3 col-xxl-4 offset-xxl-4 mb-3 mt-3">
     <div class="card-body bg-light">
         <div class="row text-center">
             <p class="lead mb-1">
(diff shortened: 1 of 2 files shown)


corylown (changes requested):
Looks good overall. The header and the info box look a bit crowded to me. Maybe add some space between them?
Screen Shot 2022-05-12 at 4 37 29 PM

ggeisler:
👍 `mt-3` on the info box element might be about right.

marlo-longley:
Updated :)

Screen Shot 2022-05-12 at 15 55 04

ggeisler:
I don't think this is the spacing Cory or I were thinking of. I believe the concern was spacing between the page title ("File: pod....") and the top of the box. See how there is little vertical whitespace between those elements in Cory's screenshot.

My suggestion was to put the `mt-3` on the outer element of the box, the one with the the `card` class. That should give us 1rem of whitespace between the page title and the start of the box. (And I can worry about adjusting it more later if necessary after this is merged.)

marlo-longley:
@ggeisler ah thank you for clarifying. I added mt-3 !

Issue #602, opened 2022-04-14
Generate an inter-stream delete list for a stream


corylown:
Given work on https://github.com/pod4lib/aggregator/pull/619 it likely makes sense to start by generating interstream  delete files (delete text file & oai) that contain records that were included in the previous stream but are not included in the current stream.

~~Blocked by #599~~

This is likely to be a slow process and may need to be calculated as a background job and stored in a file ~~or the database for future reference:~~

~~GenerateInterStreamDeleteJob: Given a current or former stream generate a list of MARC record 001 IDs that are present in the previous default stream and absent in the given stream. Store this list in a file or db table associated with the given stream (or associated with the stream's history if we allow streams to become a default more than once).~~

corylown:
Closed via #714

Issue #601, opened 2022-04-14
Generate an inter-stream delta for a stream


corylown:
Given work on #619 it likely makes sense to start by generating interstream delta files (marc21, marcxml & oai) that contain records that have changed since the previous stream.

~~Blocked by #599~~

This is likely to be a slow process and may need to be calculated as a background job and stored in a file ~~or the database for future reference:~~

~~GenerateInterStreamDeltaJob: Given a current or former default stream generate a list of MARC record 001 IDs for records that have NOT changed between the given stream and the prior default stream (or streams, depending on the method employed). Store this list in a file or db table associated with the given stream (or associated with the stream's history if we allow streams to become a default more than once).~~

JohnMarkOckerbloom:
This may be a convenient data structure to generate for efficient implementation of incremental harvesting.  It's not strictly necessary, however, and can be dropped if it doesn't make implementation easier or more efficient.  (The developers can tell whether it does more easily than I can.)  What an incremental OAI harvest needs to return are the records that *have* changed (including information on those that have been deleted, which is why inter-stream deletes are important to calculate), rather than the records that have not.

corylown:
Closed via #714

Issue #417, opened 2022-02-02
Development of alternative pathways to harvest POD data, e.g. OAI-PMH or FTP


corylown:

anarchivist:
Per convos with John, James and Sebastian at Index Data, OAI-PMH is the preferred mechanism. We're going to have  followup conversations with Index Data devs to get clearer specs.

corylown:
# Technical Notes

_from @JohnMarkOckerbloom_

Two ways for POD to provide OAI-PMH service (possibly useful background for discussions with Index Data)

There are two basic ways for POD to provide OAI-PMH exports of institutional records:

**1. providing them in stream-specific sets, or
2. providing them in organization-specific sets.**

The former is easier to implement on the server side, but is more burdensome for clients like ReShare to cope with. The latter is easier for clients to consume, but is more challenging for POD to implement and support.

Both are potentially feasible to work with, though. (It's also possible, though probably not necessary, to implement both kinds of sets.)  Below I have some notes on how servers and clients can use them, and on some of the tradeoffs involved.  (Let me know if you'd like this in a fancier document format, or placed somewhere for reference.)

## 1. Stream-specific sets

If POD provides separate sets for each stream, then it can easily provide both full and incremental exports.  A request for records will simply iterate through the normalized data files in a stream in reverse chronological order, skipping records with IDs already dealt with, and serving records with IDs not seen yet.  If from and/or until bounds are used, only the files that were created within those bounds are iterated through. This method supports both full and incremental harvesting of streams.

However, ReShare and most clients are most interested in keeping up to date with records from an organization, and not necessarily from a particular stream.  If an organization designates a new default stream, and starts uploading files to it, they will not be seen by a client that's expecting new files to be in the old stream.  Moreover, even if the client knows to switch to the new strem, it will have to start over with a full harvest, even if only a small number of records have changed since the last stream.

Mitigating these problems are the possibilities for a server to inform the client about the new default stream, and for organizations to avoid switching to new streams too often. While there is no standard for doing so, it is possible for a server and clients to agree on a convention for indicating default streams for each organization through the OAI-PMH protocol.  For example, the set elements returned in response to a ListSets request could include a setDescription element that indicates whether the set is for a current default stream from a particular organization.  A client that starts with a ListSets request could thus learn which sets it should harvest to get current records from an organization.  And if new default streams are not created too often, the client might not find the need to harvest full record sets excessively burdensome.

## 2. Organization-specific sets

[...]

PR #832, opened 2022-05-24
OAI-PMH ListRecords and friends

thatbudakguy:
# changelog
- fills out the tests for the entire OAI implementation, including ListRecords
- implements resumptionTokens for OAI requests
- refactors ListRecords to use stream-based sets
- handles a few more error/edge cases for assorted other OAI request types
- generates OAI-XML files as part of full and delta dumps that are used to page thru records in OAI requests
- fixes a bug related to generating empty OAI-XML files, and tests for it
- updates the naming of delta dump files to include more information, as we do for full dumps
- adds/makes some fixtures easier to read and fixes some misc. typos/copy-paste errors around fixtures.
# rationale
1. using streams as sets avoids the thorny problem of inter-stream deltas (see e.g. #601, #602) and we have heard from IndexData that this is acceptable for the time being (see conversation on #417). our `ListSets` response includes some metadata in the `setDescription` block intended to be parsed by IndexData (if necessary) to indicate whether a set is a current or former default stream, and the dates for which it was/is the default.
2. IndexData has indicated that processing records in blocks of about 100 at a time is reasonable. the simplest implementation for paged OAI responses is to pregenerate the OAI-XML ahead of time as part of the dump jobs, in 100-record-size "pages", and then serve them in a paginated manner from the controller. not every page will include the full 100 records, but no page will include more than 100 records, and progressing through the pages via `resumptionToken`s will eventually yield the full set, which is OAI-compliant.
3. a `resumptionToken` is a Base64-urlsafe-encoded combination of `set` (see 1. above), `page` (see 2. above), `from_date` (corresponding to the OAI `from` param), and `until_date` (corresponding to the OAI `until` param), any of which may be `nil`. the token is used to construct a filtered set of `NormalizedDump`s, which is mapped to a list of OAI-XML "pages", and then indexed using `page`. tokens have a read-only `version` specifier to ensure that, if this API changes, tokens encoded using a former version of the API will be rendered invalid.
4. to ensure uniqueness, OAI record IDs for MARC records are currently defined via [OAI-ID guidelines](http://www.openarchives.org/OAI/2.0/guidelines-oai-identifier.htm) as `oai:pod.stanford.edu:[org-slug]:[stream-id]:[marc001]`.

### DIFF
--- a/app/controllers/concerns/oai_concern.rb
+++ b/app/controllers/concerns/oai_concern.rb
@@ -57,19 +57,12 @@ module OaiConcern
   class ResumptionToken
     def self.encode(set: nil, page: nil, from_date: nil, until_date: nil)
       validate(set, page, from_date, until_date)
-
       Base64.urlsafe_encode64([set, page, from_date, until_date].join(';'))
     end
 
     def self.decode(token)
       set, page, from_date, until_date = Base64.urlsafe_decode64(token).split(';')
-
-      begin
-        validate(set, page, from_date, until_date)
-      rescue ArgumentError
-        raise OaiConcern::BadResumptionToken
-      end
-
+      validate(set, page, from_date, until_date)
       [set, page, from_date, until_date]
     end
 
@@ -78,6 +71,8 @@ module OaiConcern
       Integer(page) if page.present?
       Date.parse(from_date) if from_date.present?
       Date.parse(until_date) if until_date.present?
+    rescue ArgumentError
+      raise OaiConcern::BadResumptionToken
     end
   end
 
--- a/app/controllers/concerns/oai_concern.rb
+++ b/app/controllers/concerns/oai_concern.rb
@@ -60,7 +60,22 @@ module OaiConcern
     end
 
     def self.decode(token)
-      Base64.urlsafe_decode64(token).split(';')
+      set, page, from_date, until_date = Base64.urlsafe_decode64(token).split(';')
+
+      begin
+        validate(set, page, from_date, until_date)
+      rescue ArgumentError
+        raise OaiConcern::BadResumptionToken
+      end
+
+      [set, page, from_date, until_date]
+    end
+
+    def self.validate(set, page, from_date, until_date)
+      Integer(set) if set.present?
+      Integer(page) if page.present?
+      Date.parse(from_date) if from_date.present?
+      Date.parse(until_date) if until_date.present?
     end
   end
 
--- a/app/controllers/concerns/oai_concern.rb
+++ b/app/controllers/concerns/oai_concern.rb
@@ -56,11 +56,13 @@ module OaiConcern
   # list.
   class ResumptionToken
     def self.encode(set, page, from_date, until_date)
-      Base64.urlsafe_encode64([set, page, from_date, until_date].join(';'))
+      # Base64.urlsafe_encode64([set, page, from_date, until_date].join(';'))
+      [set, page, from_date, until_date].join('|')
     end
 
     def self.decode(token)
-      Base64.urlsafe_decode64(token).split(';')
+      # Base64.urlsafe_decode64(token).split(';')
+      token.split('|')
     end
   end
 
--- a/app/controllers/concerns/oai_concern.rb
+++ b/app/controllers/concerns/oai_concern.rb
@@ -55,7 +55,9 @@ module OaiConcern
   # A token lets you construct a list of records and point to somewhere in that
   # list.
   class ResumptionToken
-    def self.encode(set, page, from_date, until_date)
+    def self.encode(set: nil, page: nil, from_date: nil, until_date: nil)
+      validate(set, page, from_date, until_date)
+
       Base64.urlsafe_encode64([set, page, from_date, until_date].join(';'))
     end
 
--- a/app/controllers/concerns/oai_concern.rb
+++ b/app/controllers/concerns/oai_concern.rb
@@ -55,25 +55,48 @@ module OaiConcern
   # A token lets you construct a list of records and point to somewhere in that
   # list.
   class ResumptionToken
-    def self.encode(set: nil, page: nil, from_date: nil, until_date: nil)
-      validate(set, page, from_date, until_date)
-      Base64.urlsafe_encode64([set, page, from_date, until_date].join(';'))
+    attr_reader :set, :page, :from_date, :until_date, :version
+
+    def initialize(set: nil, page: nil, from_date: nil, until_date: nil)
+      @set = set
+      @page = page
+      @from_date = from_date
+      @until_date = until_date
+      @version = ResumptionToken.version
+      raise BadResumptionToken unless valid?
     end
 
-    def self.decode(token)
-      set, page, from_date, until_date = Base64.urlsafe_decode64(token).split(';')
-      validate(set, page, from_date, until_date)
-      [set, page, from_date, until_date]
+    # Error if the token is invalid or if it is a version other than ours
+    def self.decode(string)
+      set, page, from_date, until_date, version = Base64.urlsafe_decode64(string).split(';')
+      raise BadResumptionToken unless version == ResumptionToken.version
+
+      token = ResumptionToken.new(set: set, page: page, from_date: from_date, until_date: until_date)
+      raise BadResumptionToken unless token.valid?
+
+      token
     end
 
-    def self.validate(set, page, from_date, until_date)
-      Integer(set) if set.present?
-      Integer(page) if page.present?
-      Date.parse(from_date) if from_date.present?
-      Date.parse(until_date) if until_date.present?
+    def self.version
+      'v1.0'
+    end
+
+    def encode
+      Base64.urlsafe_encode64([@set, @page, @from_date, @until_date, @version].join(';'))
+    end
+
+    # rubocop:disable Metrics/AbcSize
+    # valid iff all values can be parsed and set/page are nonnegative integers
+    def valid?
+      Integer(set) if @set.present?
+      Integer(page) if @page.present?
+      Date.parse(from_date) if @from_date.present?
+      Date.parse(until_date) if @until_date.present?
+      !set.to_i.negative? && !page.to_i.negative?
     rescue ArgumentError
-      raise OaiConcern::BadResumptionToken
+      false
     end
+    # rubocop:enable Metrics/AbcSize
   end
 
   # rubocop:disable Metrics/BlockLength
--- a/app/controllers/concerns/oai_concern.rb
+++ b/app/controllers/concerns/oai_concern.rb
@@ -99,7 +99,6 @@ module OaiConcern
     # rubocop:enable Metrics/AbcSize
   end
 
-  # rubocop:disable Metrics/BlockLength
   included do
     private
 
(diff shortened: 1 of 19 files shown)


app/controllers/concerns/oai_concern.rb
```diff
@@ -51,7 +51,35 @@ def message
     end
   end
 
+  # Token for requesting records: base64-encoded combo of filters & page cursor
+  # A token lets you construct a list of records and point to somewhere in that
+  # list.
+  class ResumptionToken
+    def self.encode(set: nil, page: nil, from_date: nil, until_date: nil)
+      validate(set, page, from_date, until_date)
+      Base64.urlsafe_encode64([set, page, from_date, until_date].join(';'))
```
cbeer, line 60: I'd suggest we add some kind of version string (e.g. `v1`)  at the start of this, in case we need to change the format in the future.

app/controllers/concerns/oai_concern.rb
```diff
@@ -61,5 +89,38 @@ def oai_xmlns
...
+    def oai_dc_xmlns
+      {
+        'xmlns:oai_dc' => 'http://www.openarchives.org/OAI/2.0/oai_dc/',
+        'xmlns:dc' => 'http://purl.org/dc/elements/1.1/',
+        'xmlns:xsi' => 'http://www.w3.org/2001/XMLSchema-instance',
+        'xsi:schemaLocation' => 'http://www.openarchives.org/OAI/2.0/oai_dc/ http://www.openarchives.org/OAI/2.0/oai_dc.xsd'
+      }
+    end
+
+    def list_sets_description(stream)
```
cbeer, line 105: I'm not sure if we have any other use for it, but I'm finding this a little hard to read.

I wonder if this would be easier to write as `Stream#oai_description`? Or, maybe just more compactly as something like:

```
def list_sets_description(stream)
  "#{stream.default? ? 'Current' : 'Former'} default stream for #{stream.organization.slug} #{list_sets_date_ranges.join(', ')"
end
```

app/controllers/oai_controller.rb
```diff
@@ -93,23 +108,62 @@ def identify_params
...
   end
 
-  # rubocop:disable Metrics/AbcSize
-  def list_record_normalized_dump_candidates
-    @list_record_normalized_dump_candidates ||= begin
-      current_full_dump = @stream.current_full_dump
+  # Get a page of OAI-XML records and a token pointing to the next page
+  def next_record_page(token = nil)
+    # parse the token if we were provided one
+    set, page, from_date, until_date = OaiConcern::ResumptionToken.decode(token) if token
```
cbeer, line 114: It's a little funny we're encoding a resumption token (in the caller) in order to immediately decode it (here). Maybe `OaiConcern::ResumptionToken` would be easier to work with if we made it an actual class with attributes instead of a container for class-level methods? Maybe something like:

```
class OaiResumptionToken
  attr_reader :set, :page, :from_date, :until_date

  def initialize; end
  def encode; end
  def valid?; end
end
```

app/controllers/oai_controller.rb
```diff
@@ -93,23 +108,62 @@ def identify_params
...
+    page = page.to_i
+
+    # filter normalized dumps and get the corresponding OAI-XML pages
+    # NOTE: each page is guaranteed to have < OAIPMHWriter::max_records_per_file,
+    # but some pages will have exactly that number and others won't. The sequence
+    # isn't predictable; the only thing we promise is that all pages are less than
+    # that size.
+    pages = normalized_dumps(set, from_date, until_date).flat_map(&:oai_xml_attachments)
```
cbeer, line 122: I'm eyeing this with a little concern for our larger dumps. Is it worth getting ActiveRecord to do some of this pagination for us, or is it fine in practice?

app/controllers/oai_controller.rb
```diff
@@ -93,23 +108,62 @@ def identify_params
...
+    dumps
+  end
 
-      dumps
-    end
+  def filter_dumps(streams, from_date, until_date)
+    # get candidate dumps (the current full dump and its deltas for each stream)
+    dumps = streams.flat_map(&:current_dumps).sort_by(&:created_at)
+
+    # filter candidate dumps (by from date and until date)
```
cbeer, line 161: Could these `select`s end up as part of the scoping for current_dumps?

cbeer:
Could we update the PR description to include a little bit of "why" for these changes? I think it's maybe captured in some of the tickets that culminated in this, but to be kind to our future selves maybe we can include it here. I'm thinking things like:

- set per stream instead of set per org
-  100 OAI records per file
- OAI record id uniqueness
- etc

app/controllers/oai_controller.rb
thatbudakguy: Definitely in favor of pushing more stuff onto the database if possible. How do we do that in ActiveRecord? Is that a Kaminari thing?

app/jobs/generate_delta_dump_job.rb
```diff
@@ -28,27 +28,38 @@ def perform(organization)
     progress.total = uploads.sum(&:marc_records_count)
 
     delta_dump = full_dump.deltas.create(stream_id: full_dump.stream_id)
-
-    writer = MarcRecordWriterService.new("#{organization.slug}-#{Time.zone.today}-delta")
+    base_name = "#{organization.slug}-#{Time.zone.today}-delta"
+    writer = MarcRecordWriterService.new(base_name)
+    oai_file_counter = 0
```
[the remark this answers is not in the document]
thatbudakguy: I think so; I ended up touching this in my other branch that factors out the common code across the two jobs and that's one of the changes in it.

marlo-longley:
One broader question I had about this PR was about documenting the API, its routes, params, and responses, and where the appropriate place for that will be. It seems like the big `oai_spec` file describes the API really well, reading it was a good baseline for me to get oriented. If there's work around documentation to be done in a separate PR I am happy to help with that.

corylown:
@marlo-longley that's a good question. @JohnMarkOckerbloom is working on documenting the API in the wiki: https://github.com/pod4lib/aggregator/wiki/Harvesting-using-OAI-PMH

app/controllers/concerns/oai_concern.rb
thatbudakguy: Great suggestion; added

app/controllers/oai_controller.rb
thatbudakguy: Implemented this and pushed a little bit more error handling on this class.

thatbudakguy:
added a short list of major decisions & their rationale to the top comment.

app/controllers/concerns/oai_concern.rb
thatbudakguy: Moved this to `Stream` and tweaked our `ListSets` response slightly so that the dates a stream was default and its organization slug are a bit more machine-parseable (moved them to dublin core fields within the description). Also added tests for the dublin core fields. cc @corylown

app/controllers/oai_controller.rb
thatbudakguy: Would that just involve adding optional arguments to `current_dumps`, like:
```rb
  def current_dumps(from_date, until_date)
...
  end
```

If so, I'm not sure how to call it in a way such that I can still use ActiveRecord `select`, since `current_dumps` on its own just returns an array. I think I need a little more explanation to implement it, but I'm on board with it.

app/jobs/generate_delta_dump_job.rb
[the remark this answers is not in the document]
thatbudakguy: Going to leave this as-is for now and address when refactoring the two jobs.

app/controllers/concerns/oai_concern.rb
thatbudakguy: pic for reference:

Screen Shot 2022-05-25 at 15 48 33

app/controllers/oai_controller.rb
corylown: @thatbudakguy I think `current_dumps` could be re-written using ActiveRecord. Maybe there's a more clever way to do this, but one way would be to first get the id of the most recent full dump and then write a query with `or` to fetch the full dump and its deltas. You can also build  up a query with conditional clauses (such as for filtering by from/until dates, and maybe to support paging too).:

```ruby
full_dump = normalized_dumps.full_dumps.order(created_at: :desc).limit(1)
full_dump_id = full_dump.first.id
dumps_query = NormalizedDump.where(id: full_dump_id).or(NormalizedDump.where(full_dump_id: full_dump_id)).order(created_at: :asc)
if from_date && until_date
  dumps_query = dumps_query.where('created_at BETWEEN ? AND ?', Time.zone.parse(from_date).beginning_of_day, Time.zone.parse(until_date).end_of_day)
else
.... etc.
end
dumps_query
```

I think the paging could be handled here too, by applying `limit(1)` and `offset(page)` to the query.

JohnMarkOckerbloom:
I'd said "up to 1000 records" rather than 100 in the documentation, based on what Index Data said they could handle.  If it's not a burden either for coding or for the performance to make the chunk size larger, a larger record chunk size would cut down on the number of back and forth calls to the server.  (Based on the responses I've seen from HathiTrust, 500 records might be a sweet spot for a substantial number of records that's still a manageable response size.  That would still make for 2,000 transactions needed to harvest 1 million records, but that's probably better than 10,000.)

JohnMarkOckerbloom:
It would also be good to be very unambiguous about indicating default sets in a machine-readable manner. (I see "default" is in the description element of the setDescription in the example, but it's in the middle of other free text.)

Strictly speaking, the setDescription does not need to be Dublin Core (though it's recommended as good practice, and I suppose it saves the trouble of declaring a different XML schema).   If we do want to keep using Dublin Core, though, we could have a description element saying just "default", if applicable.  Between that and putting the provider institution slug in "publisher", that should give a machine processor enough information to automatically know what to do with the sets.  (If we also want a human-readable description, it's also acceptable to include multiple description elements, as long as it's clear to a machine processor what to watch for.)

thatbudakguy:
@corylown do you remember how we ended up with 100? should we change `Settings.oai_max_page_size`?

corylown:
@thatbudakguy I think we ended up with 100 because that was the processing chunk size before we implemented OAI-PMH. We could try bumping that up to 500 and see if we run into memory issues.

JohnMarkOckerbloom:
Thanks!  If it does cause memory issues or other performance problems, it's okay as far as I know to leave it at 100.  (I'll let folks know if I hear otherwise from clients.)  But I thought I'd note the advantages of larger chunk sizes if it was just an arbitrary constant.

thatbudakguy:
@JohnMarkOckerbloom good point re: the description element. I'm updating it to look like the following; in production the `setName` is often something more useful like `prod`, which is what Columbia called their default stream. I put the default status in `type` so it's on its own, and switched `publisher` to `contributor` which seemed a little more appropriate.

Screen Shot 2022-05-26 at 11 43 42

app/controllers/oai_controller.rb
corylown: @thatbudakguy I was messing around with this also and reaching similar conclusions.

corylown: I think filtering the ActiveStorage::Attachment objects may be feasible if we start with the list of relevant NormalizedDump ids: `record_id` in the query:

```
ActiveStorage::Attachment.where(record_type: 'NormalizedDump', name: 'oai_xml', record_id: [60,62]).order(crea
ted_at: :asc).limit(1).offset(2)
```

corylown (approved):

Issue #672, opened 2022-04-22
Add Activity summary tables to Activity page


ggeisler:
This is a story ticket intended to keep track of a set of related work tickets, plus provide an overview of how this tickets fit into the current Activity page.

- [x] #673
- [x] #674
- [x] #675
- [x] #676
- [ ] #677
- [x] #678
- [x] #679

## Context
We currently have an Activity page, visible only to POD admins. This story involves adding the section labeled "A" in the mockup below to the existing Activity page, which currently contains the legend and table in the section labeled "B" in the mockup (the "Filter uploads" box in section "B" is described in #629).

<img width="979" alt="context" src="https://user-images.githubusercontent.com/101482/164817431-89cf7594-fd23-4b83-bc9d-3aacb711ec29.png">

ktamaral:
Branch in progress: https://github.com/pod4lib/aggregator/tree/activity-summary

JohnMarkOckerbloom:
I've gone through the various tickets, and while I don't have a lot to say about them at this point, they look appealing and useful!  We may well learn from experience how to adjust these for the needs of admins, but this looks like a good start.

marlo-longley:
@ggeisler Quick question about this layout - is section B supposed to show up on every tab view? Or only the uploads tab. My working assumption has been section B will show up regardless of which tab is active. But realized I should check with you. Thanks!

ggeisler:
@marlo-longley Yes, section B should always be there. That screenshot is to provide context for the new section A, just showing how it appears before section B on the page.

marlo-longley:
Hi @ggeisler - wanted to get your input after a conversation I had with @corylown this morning.

We've been experimenting with different caching strategies because this page is very slow to load. Specifically, loading and paging through the uploads list (section B in your mockup) is extremely slow. We realized that caching won't work well here because it will cause data to be not up to date / live / accurate. So we want to avoid caching after all.

So we were wondering if there is a way to separate section B out from section A, so that they are actually viewed on different pages, reducing load time. There's some more context in the discussion in this PR https://github.com/pod4lib/aggregator/pull/841. This is something we can maybe chat about after standup on Monday. Thanks!

PR #833, opened 2022-05-24
Activity summary jobs

### LINKED ISSUE #675: Add "Job status" table to Activity summary section
Part of #672.

Below is a suggested start for the Job status tab. As with all of the Activity summary tables, the main goal is to start with something, and we can refine and adjust based on what data is actually feasible to gather and what ideas we might get for showing more useful things.

job-status

---

- [x] **A:** `Provider` column. Link to that provider's current stream > Processing status tab.
- [x] **B:** `Active jobs` column. Display the count of active jobs for that provider. If more than zero, link to the provider page > Processing status tab. Right-align this column.
- [x] **C:** `Last active job started` column. Date/time of the last job started. Blank if there are no active jobs. (I guess it could link to the provider page > Processing status tab, since we know if there is a timestamp it means there is at least one active job and that section won't be empty?)
- [x] **D:** `Jobs needing attention` column. Display the count of jobs needing attention for that provider. If more than zero, link to the provider page > Processing status tab. Right-align this column.

marlo-longley:
Continues work on #672

Closes #675


thatbudakguy (approved):
looks good to me; I like the decision to split out partials for each of the tabs.

corylown (approved):

PR #841, opened 2022-05-26
Add Normalized data and Users tabs to Activity summary

### LINKED ISSUE #676: Add "Normalized data" table to Activity summary section
Part of #672.

Below is a suggested start for the Normalized data tab. As with all of the Activity summary tables, the main goal is to start with something, and we can refine and adjust based on what data is actually feasible to gather and what ideas we might get for showing more useful things.

normalized-data

---

- [x] **A:** `Provider` column. Link to that provider's page > Normalized data tab.
- [x] **B:** `Last full dump` column. Date/time of the last full dump from that provider. If that provider doesn't have a full dump, display the `exclamation-triangle-fill` icon colored warning yellow.
- [x] **C:** `Dump records` column. Count of the number of records in the last full dump. Zero if they don't have a full dump. Right-align this column.
- [x] **D:** `Last delta` column. Date/time of the last delta file from that provider. If that provider doesn't have a delta file, display the `exclamation-triangle-fill` icon colored warning yellow.
- [x] **E:** `Last deletes` column. Date/time of the last deletes file from that provider. If that provider doesn't have a deletes file, display the `exclamation-triangle-fill` icon colored warning yellow.

### LINKED ISSUE #678: Add "Users" table to Activity summary section
Part of #672.

Below is a suggested start for the Users tab. As with all of the Activity summary tables, the main goal is to start with something, and we can refine and adjust based on what data is actually feasible to gather and what ideas we might get for showing more useful things.

users

---

- [x] **A:** `Provider` column. Link to that provider's Manage organization > Users tab.
- [x] **B:** The Members, Owners, and POD admin columns are just counts of the number of users in each role for that organization. Right-align these columns.

marlo-longley:
Also refactor into Dashboard class with caching

Closes #676 
Closes #678


app/helpers/dashboard_helper.rb
```diff
@@ -41,4 +41,33 @@ def any_successes?(statuses)
...
+
+  def dump_record_count(dump)
+    0 if dump.nil?
+
+    if dump.marcxml&.attachment && dump.marcxml.attachment.metadata
+      # Use marcxml as the source of the record count. marc21 count may differ.
+      dump.marcxml.attachment.metadata['count']
+    else
+      0
+    end
```
thatbudakguy, line 53: I'd love to see this as a method on `NormalizedDump`; I can see using it some other places (like in the OAI controller) too.

app/helpers/dashboard_helper.rb
```diff
@@ -41,4 +41,33 @@ def any_successes?(statuses)
...
+    }
+  end
+
+  def highest_role(user)
+    roles = user.roles.map(&:name).uniq
+
+    return :admin if roles.include? 'admin'
+    return :owner if roles.include? 'owner'
+    return :member if roles.include? 'member'
+  end
```
thatbudakguy, line 72: Same thing here — this could be a method on `User`, although it's harder to immediately come up with other places we'd use it.

app/models/dashboard.rb
```diff
@@ -0,0 +1,28 @@
...
+  def recent_uploads_by_provider
+    Rails.cache.fetch(:recent_uploads_by_provider, expires_in: 1.hour) do
+      Upload.recent.where('created_at > ?', 30.days.ago).group_by(&:organization)
+    end
+  end
+
+  def job_status_groups_by_provider
+    Rails.cache.fetch(:job_status_groups_by_provider, expires_in: 1.hour) do
+      Organization.providers.index_with { |org| org.default_stream.job_tracker_status_groups }
+    end
```
thatbudakguy, line 14: Does the 1 hour cache time here mean that e.g. jobs that complete within the 1 hour window won't show as completed until 1 hour has elapsed? Not sure how the caching works.

thatbudakguy (approved):

app/models/dashboard.rb
corylown: I think this query could either not be cached (it's not that slow) or it could be cached very briefly, maybe a minute or something.

thatbudakguy: Yeah, with stuff like this I worry that people will be looking to this dashboard for up-to-the-minute information, and then be confused or start to lose trust in it when they find that it doesn't actually represent the system state sometimes (and by "people" I mean me).

marlo-longley: @thatbudakguy @corylown I had this same concern about caching vs. up to date info -- we also have some info in a different tab that is based on the Statistics job, so I worry about having 2-3 multiple different time scales for data on this page.

corylown: I agree that it's probably okay if the page is slow initially. The problem in my mind is the paging presented with the uploads list -- it's unbearably slow to wait for the data to load for each page. I wonder if this points to the need for a different design, where the uploads appear on their own page where the tabbed data isn't being loaded.

marlo-longley: See comment in https://github.com/pod4lib/aggregator/issues/672

marlo-longley:
@thatbudakguy just pushed a commit that moves the methods into their respective models like you suggested, thanks!

This push also:
Fixes #843 
Fixes #844

app/views/dashboard/_tab_normalized_data.html.erb
```diff
@@ -1,3 +1,43 @@
...
+      <thead>
+        <tr>
+          <th><%= t('dashboard.summary.table.provider') %></th>
+          <th><%= t('dashboard.summary.table.last_full_dump') %></th>
+          <th class="text-end pe-4"><%= t('dashboard.summary.table.dump_records') %></th>
+          <th><%= t('dashboard.summary.table.last_delta') %></th>
+        </tr>
+      </thead>
+      <tbody>
+        <% @dashboard.normalized_data_by_provider.sort_by{ |k, _v| _v.updated_at}.reverse.each do |provider, dump| %>
```
corylown, line 12: This doesn't account for the possibility that `_v` can be nil.

The variable name prefixed with an underscore signals that it is not used, so I think you want `sort_by { |_k, v| v.updated_at }`

app/views/dashboard/_tab_normalized_data.html.erb
```diff
@@ -1,3 +1,43 @@
...
+        <tr>
+          <td><%= link_to provider.name, normalized_data_organization_stream_path(provider, provider.default_stream) %></td>
+          <!-- Last full dump: -->
+          <td>
+              <% if dump %>
+                <%= local_time(dump.last_full_dump_at, format: datetime_display_format()) %>
```
corylown, line 18: I'm seeing `undefined method 'to_time' for nil:NilClass` from this line.

corylown (changes requested):

app/views/dashboard/_tab_normalized_data.html.erb
marlo-longley: @corylown my bad. Thanks.

marlo-longley: @corylown lol I rolled out `Time.at(0)` in my attempt to solve this nil issue in latest commit, not sure of another better way!

corylown (approved):

PR #851, opened 2022-05-31
oai identify response

thatbudakguy:
- Extract OAI repository ID to settings
- Update OAI Identify response

The validator at http://oval.base-search.net/ seemed to not like the order in which we presented our elements in the `Identify` response. Hard to imagine this is in violation of the spec, but I reordered them and also added a `description` as recommended by the spec that illustrates how we are generating OAI IDs for records. The base OAI ID for the whole aggregator is now controlled by a config setting.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -54,3 +54,6 @@ metadata_status:
 
 # max. MARC records per OAI-XML file; also max. size of one page in ListRecords
 oai_max_page_size: 500
+
+# base ID of the repository used to generate OAI IDs for records
+oai_repository_id: 'pod.stanford.edu'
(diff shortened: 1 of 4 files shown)


config/settings.yml
```diff
@@ -54,3 +54,6 @@ metadata_status:
 
 # max. MARC records per OAI-XML file; also max. size of one page in ListRecords
 oai_max_page_size: 500
+
+# base ID of the repository used to generate OAI IDs for records
+oai_repository_id: 'pod.stanford.edu'
```
corylown, line 59: More a thought than a blocking issue, but maybe we should nest these?:

```
oai:
  max_page_size: 500
  repository_id: 'pod.stanford.edu'
```

corylown (approved):

config/settings.yml
thatbudakguy: yeah, i was thinking the same — there'll probably be more going forward

PR #857, opened 2022-06-02
Lean on ActiveRecord for oai ListRecords response file and next page info.

corylown:


thatbudakguy (approved):
looks good; I like the choice to extract next_page_token too.

PR #858, opened 2022-06-02
Add table for consumer-only orgs to Providers page

### LINKED ISSUE #811: Make it easier to find Consumer-only org(s)
While the provider page reorganization makes the providers table more accurate (it doesn't include consumer-only orgs who don't have the data presented in the table), it does present a problem in that there is currently no obvious place to find consumer-only orgs (we only have one, but maybe there will be more in the future). 

We need a way to get to a list of consumer-only organizations without cluttering the UI with a link for a seldom-needed action (given there is only one consumer-only org).

marlo-longley:
Closes #811

This PR implements Option B laid out in the ticket which seemed to have been agreed upon in the discussion.

Screen Shot 2022-06-02 at 16 02 57

### DIFF
--- a/app/views/organizations/index.html.erb
+++ b/app/views/organizations/index.html.erb
@@ -15,43 +15,46 @@
     </div>
   <% end %>
 
-  <h2>Consumer-only organizations</h2>
-  <table class="table table-striped organizations mb-5 mt-4" id="consumers">
-    <thead>
-      <tr>
-        <th colspan="2">Consumer</th>
-        <th class="text-end">Users</th>
-        <th>POD contact</th>
-        <th>Actions</th>
-        <th colspan="2"></th>
-      </tr>
-    </thead>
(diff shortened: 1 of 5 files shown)


thatbudakguy (approved):
looks good!

app/views/organizations/index.html.erb
```diff
@@ -15,12 +15,52 @@
     </div>
   <% end %>
 
-  <h2>Default stream statistics by provider</h2>
+  <!-- Only show consumer-only table to admins -->
+  <% if current_user && current_user.has_role?(:admin) %>
```
corylown, line 19: Maybe this should only show up if there are `@consumers`?

marlo-longley: @corylown added a commit to address this.

corylown (approved):

PR #861, opened 2022-06-03
Replace the stream summary box with more responsive code

ggeisler:
The summary box on the Provider home pages is a little awkward in that it's hard to get the ideal width of the box at different viewports widths. It also isn't great visually: it takes up more vertical height than it really needs to.

Screen Shot 2022-06-03 at 3 38 12 PM

This PR replaces the existing code, which is based on Bootstrap's Card component, and uses CSS Grid instead. The CSS is kind of verbose but this enables us to display a box in a consistent width at desktop viewports, and break the rows of metadata into new lines when at small viewports. This approach is also more compact and I think gives a stronger emphasis to the box.

## After: Desktop
Screen Shot 2022-06-03 at 3 36 41 PM

---

## After: Small mobile
(Nothing looks great here, but presumably users would not be using POD on a phone on any kind of regular basis...)

Screen Shot 2022-06-03 at 3 37 27 PM

---

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -18,6 +18,7 @@
 @import 'pod-icons';
 @import 'rails_bootstrap_forms';
 @import 'organizations';
+@import 'streams';
 
 // background image for homepage banner
 #hero .bg-image {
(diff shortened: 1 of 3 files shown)


marlo-longley:
@ggeisler nice! I can always stand to get handier with CSS grid. I checked out the branch looks good.

marlo-longley:
@ggeisler nice! I can always stand to get handier with CSS grid. I checked out the branch looks good.

PR #864, opened 2022-06-06
Show all orgs on Activity page Uploads tab regardless of recent activity

### LINKED ISSUE #818: Uploads tab of Activity page should display uploads summary for all orgs
Relates to https://github.com/pod4lib/aggregator/issues/674 that was recently implemented.

Chicago is notably missing from the table on this page https://pod.stanford.edu/dashboard/uploads because their most recent upload is outside the 30 day window of the current logic. I think we want to show a summary for all orgs regardless of when their last upload was (in fact, it seems important and useful to show that an org has not uploaded in a while).

Change the logic for fetching upload summary info to include uploads for orgs beyond the current 30 day window.

There may be performance issues to address here as well if we expand the query to fetch data beyond 30 days ago.

marlo-longley:
Closes #818

### DIFF
--- a/app/models/dashboard.rb
+++ b/app/models/dashboard.rb
@@ -3,7 +3,25 @@
 # Summary of recent activity across the whole app
 class Dashboard
   def recent_uploads_by_provider
-    Upload.recent.where('created_at > ?', 30.days.ago).group_by(&:organization)
+    recent_uploads_by_provider = Upload.recent.where('created_at > ?', 30.days.ago).group_by(&:organization)
+    # Find any providers that didn't have uploads in the last 30 days
+    recent_inactive_orgs = Organization.providers - recent_uploads_by_provider.keys
+
+    # Add inactive providers into the returned hash
+    # Populate their last upload with upload outside the 30 day window, if one exists
+    # Otherise, if no uploads ever, set value to nil
+    # Also set a recently_inactive flag to true for use in the display
+    recent_inactive_orgs.each do |org|
+      org.recently_inactive = true
+      recent_uploads_by_provider.merge!({ org => inactive_org_upload_value(org) })
+    end
+
+    recent_uploads_by_provider
+  end
+
+  def inactive_org_upload_value(org)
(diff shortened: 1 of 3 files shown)


app/models/dashboard.rb
```diff
@@ -3,7 +3,25 @@
...
+    # Add inactive providers into the returned hash
+    # Populate their last upload with upload outside the 30 day window, if one exists
+    # Otherise, if no uploads ever, set value to nil
+    # Also set a recently_inactive flag to true for use in the display
+    recent_inactive_orgs.each do |org|
+      org.recently_inactive = true
+      recent_uploads_by_provider.merge!({ org => inactive_org_upload_value(org) })
+    end
+
+    recent_uploads_by_provider
```
corylown, line 19: Maybe instead of combining these into a single hash, return them as part of an array or hash to be handled separately as needed by the view. Untested, but something like:
```suggestion
    uploads_by_inactive_provider = recent_inactive_orgs.to_h { |org| [org, inactive_org_upload_value(org)] }
    { active_orgs: recent_uploads_by_provider, inactive_orgs: uploads_by_inactive_provider  }
```

app/views/dashboard/_tab_uploads.html.erb
```diff
@@ -21,25 +21,41 @@
```
corylown, line 21: If you use my suggestions, you'll need to refactor the view accordingly, perhaps by breaking out the table row to its own partial and passing each set of orgs to the new partial.

corylown (changes requested):
See inline comments for details. Happy to pair if my suggestions are unclear.

app/views/dashboard/_tab_uploads.html.erb
marlo-longley: @corylown thanks for the review! I'll go with your suggestion. I guess I thought that info might be useful somewhere else in the app at some point to know which apps were inactive. But agree we can keep it simple for now.

Will address comments this afternoon.

corylown (approved):

PR #865, opened 2022-06-08
Increase OAI page size to 1000 records

corylown:
With memory bumped to 64G it might make sense to try increasing the records/oai file limit 1000. This is IndexData's ideal per request harvest size.


thatbudakguy (approved):
It might be good to get some numbers re: response time tradeoff for this. Obviously serving more records at a time is more efficient, but it presumably trades off with time needed to send the actual HTTP response. Hopefully we don't ever reach the intersection point of that graph and dip into territory where slow responses make the overall process less efficient.

thatbudakguy:
For context, here's the timings for [a request to get Harvard's default stream in current prod](https://pod.stanford.edu/oai?verb=ListRecords&set=301&metadataPrefix=marc21). It's pretty fast (thanks to #857, I think), but if you are paging through many pages, the costs adds up.

I think we can ignore the orange and pink bars. Blue is (I'm guessing) the time the server spends doing things like ActiveRecord queries. Green is (again, guessing) the time it takes to download the actual response. If we double the OAI page size and green doesn't show a huge increase, I think we're good to go.

Screen Shot 2022-06-09 at 10 08 24

PR #869, opened 2022-06-09
Save NormalizedDump during jobs to free file handles. Add published_at timestamp when job completes.

corylown:


thatbudakguy (approved):
LGTM; interesting to read the discussion about the `Thread.pass`. hope it helps!

PR #876, opened 2022-06-13
Adjust Stream#files to return attachments to be consistent with Upload#files

### LINKED ISSUE #875: Un-normalized files listed in resourcelist resolve to the wrong files.
URLs listed in a streams resourcelist for harvesting un-normalized data resolve to the wrong files:

https://pod.stanford.edu/organizations/mit/streams/pod_stream/resourcelist
https://pod.stanford.edu/file/40702/POD_ALMA_EXPORT_20220421_220411%5B019%5D_new_9.xml

The file above should be a MARC XML file with 100,000 records submitted by MIT, instead it downloads a delete file submitted by duke.

corylown:
Closes #875

### DIFF
--- a/app/jobs/update_organization_statistics_job.rb
+++ b/app/jobs/update_organization_statistics_job.rb
@@ -46,7 +46,7 @@ class UpdateOrganizationStatisticsJob < ApplicationJob
     {
       unique_record_count: context.marc_records.distinct.count(:marc001),
       record_count: context.marc_records.size,
-      file_size: context.files.sum(:byte_size),
+      file_size: context.files.sum { |file| file.blob.byte_size },
       file_count: context.files.size
     }
   end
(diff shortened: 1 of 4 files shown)


cbeer:
Aren't we already constructing urls using the attachment id?

https://github.com/pod4lib/aggregator/blob/ebb3fd9c4c09ef3c0ed8af8103c48bcb13ff86f8/config/routes.rb#L82

corylown:
@stream.files are blobs though, right?: https://github.com/pod4lib/aggregator/blob/ebb3fd9c4c09ef3c0ed8af8103c48bcb13ff86f8/app/views/streams/show.xml.builder#L13

cbeer:
Ha. `Stream#files` are blobs, `Upload#files` are attachments.

cbeer:
Makes sense to me. `Stream#files` seems like it'll be easier to track down all the references?

PR #877, opened 2022-06-14
The oai record timestamp should be the dump create date not the record upload date.

### LINKED ISSUE #873: OAI endpoint returns unexpected records with a `datestamp` 1 day before the specified `from` date (repeated harvests)
Reported by IndexData. May require followup to determine exact nature of the issue. I have not confirmed this behavior.

corylown:
Fixes #873


thatbudakguy (approved):
this is a sane solution to me; it makes sense to consider the time that the record was actually successfully processed vs. when it was uploaded.

PR #940, opened 2023-01-24
Use consistent ENV var for redis: SIDEKIQ_REDIS_URL

corylown:
We were using both `REDIS_URL` and `SIDEKIQ_REDIS_URL` in various places. Puppet is expecting `SIDEKIQ_REDIS_URL` when we switch from using a locally running Redis to one hosted on its own VM: https://github.com/sul-dlss/puppet/blob/production/hieradata/node/pod-prod.stanford.edu.eyaml#L63-L65

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ One solution to this is to temporarily switch to using PostgreSQL and Sidekiq wi
 ```yaml
 development:
   adapter: redis
-  url: <%= ENV.fetch("REDIS_URL") { "redis://localhost:6379/1" } %>
+  url: <%= ENV.fetch("SIDEKIQ_REDIS_URL") { "redis://localhost:6379/1" } %>
   channel_prefix: aggregator_development
 ```
 
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ One solution to this is to temporarily switch to using PostgreSQL and Sidekiq wi
 ```yaml
 development:
   adapter: redis
-  url: <%= ENV.fetch("SIDEKIQ_REDIS_URL") { "redis://localhost:6379/1" } %>
+  url: <%= ENV.fetch("ACTION_CABLE_REDIS_URL") { "redis://localhost:6379/3" } %>
   channel_prefix: aggregator_development
 ```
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -62,7 +62,7 @@ One solution to this is to temporarily switch to using PostgreSQL and Sidekiq wi
 ```yaml
 development:
   adapter: redis
-  url: <%= ENV.fetch("REDIS_URL") { "redis://localhost:6379/1" } %>
+  url: <%= ENV.fetch("SIDEKIQ_REDIS_URL") { "redis://localhost:6379/1" } %>
```
jcoyne, line 65: This is for actioncable, not sidekiq.  Are you sure you want to change it?

config/cable.yml
```diff
@@ -1,7 +1,7 @@
 development:
-  <% if ENV.include? "REDIS_URL" %>
+  <% if ENV.include? "SIDEKIQ_REDIS_URL" %>
```
jcoyne, line 2: This is actioncable too.

config/environments/production.rb
```diff
@@ -56,7 +56,7 @@
   config.log_tags = [ :request_id ]
 
   # Use a different cache store in production.
-  config.cache_store = :redis_cache_store
+  config.cache_store = :redis_cache_store, { url: ENV.fetch("SIDEKIQ_REDIS_URL") { "redis://localhost:6379/1" } }
```
jcoyne, line 59: It seems to me like we should not use `SIDEKIQ_REDIS_URL` as the name for a redis instance that is used for more than Sidekiq.   For example in Exhibits we use a different Redis database for the cache than for sidekiq. https://github.com/sul-dlss/shared_configs/blob/exhibits-prod/config/initializers/sidekiq.rb#L9 (note the `/1` vs `/2` difference)

jcoyne (changes requested):

README.md
corylown: It seems like in our other apps we keep actioncable included even if it's not actively being used by the app. Maybe we should do this same here? I modified the PR to reference`ACTION_CABLE_REDIS` in `cable.yml`

jcoyne (approved):

PR #943, opened 2023-01-25
Change default redis db numbers to start at 0

corylown:
In the currently released version of aggregator the sidekiq redis db would default to `/0` so it seems better to keep that as the default redis db number for sidekiq.


thatbudakguy (approved):
needs rebase

PR #958, opened 2023-02-03
Stop calculating and displaying daily POD organization stats

corylown:
These calculations are expensive and essentially duplicate what we can already get from the stats calculated on each default stream. The only real purpose was to display the record count charts, which maybe have limited value. The overall pod record counts and unique record counts are now derived from the default stream stats, which should be roughly equivalent.


app/controllers/organizations_controller.rb
```diff
@@ -4,12 +4,6 @@
 class OrganizationsController < ApplicationController
   load_and_authorize_resource
 
-  # We need unsafe_inline for Chartkick style_src (despite the nonce for script_src)
-  # See https://github.com/ankane/chartkick/blob/master/guides/Content-Security-Policy.md
-  content_security_policy only: :index do |policy|
-    policy.style_src :self, :unsafe_inline
-  end
-
```
thatbudakguy, line 12: nice that we can ditch this, too

thatbudakguy (approved):

PR #1015, opened 2023-12-04
Update dependencies

sul-devops-team:


app/jobs/marc_profiling_job.rb
```diff
@@ -22,9 +22,11 @@ def add(item)
       @sample[j] = [@size_so_far, item] if @wanted > j
     end
 
+    # rubocop:disable Style/HashEachMethods
```
jcoyne, line 25: Why are we disabling this rather than fixing it?

jcoyne: Thank you. I will create a bug report for rubocop then.

PR #1032, opened 2024-02-07
This prevents the cron from logging the Sidekiq startup message

jcoyne:
This helps keep my inbox clean.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -50,7 +50,7 @@ Rails.application.configure do
 
   # Include generic and useful information about system operation, but avoid logging too much
   # information to avoid inadvertent exposure of personally identifiable information (PII).
-  config.log_level = :info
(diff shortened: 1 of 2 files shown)


config/environments/production.rb
```diff
@@ -50,7 +50,7 @@
 
   # Include generic and useful information about system operation, but avoid logging too much
   # information to avoid inadvertent exposure of personally identifiable information (PII).
-  config.log_level = :info
+  config.log_level = ENV.fetch("RAILS_LOG_LEVEL", "info")
```
jcoyne, line 53: This is how the new rails template does it: https://github.com/rails/rails/blob/bab4aa7cb25112846e04cea907361a1de0f126ef/railties/lib/rails/generators/rails/app/templates/config/environments/production.rb.tt#L73C1-L73C58

corylown (approved):

PR #1088, opened 2024-10-28
Switch from turbolinks and rails-ujs to turbo-rails

jcoyne:


corylown (approved):
Looks good to me. Needs rebase before merge.

PR #1185, opened 2025-10-01
Update docker setup with new Rails and image versions

corylown:


corylown:
@dnoneill could we merge this to get these changes in? I think the css not building is a problem that could be addressed as separate work.

dnoneill:
@corylown That sounds good to be. I am just going add a comment here. You will need to run `rake css:build` to be rid of the error.

dnoneill (approved):

PR #1278, opened 2025-10-16
Write MARCXML using Ox.

cbeer:
Anecdotally, this should give us at least a 2x-4x boost in full dump time 🤷‍♂️


app/services/ox_marc_xml_writer.rb
```diff
@@ -0,0 +1,89 @@
+# frozen_string_literal: true
+
+# Write MARC records to XML using Nokogiri instead of REXML (from upstream)
```
corylown: Maybe leftover from previous iteration?
```suggestion
# Write MARC records to XML using Ox instead of REXML (from upstream)
```

PR #1299, opened 2025-10-20
add more tests to accessibility tests, add role to i

### LINKED ISSUE #1261: Add logged in tests to accessibility_spec
Right now we are just checking our homepage when the user is logged out.

Probably best to wait until all the other accessibility tickets are done because this might raise of these issues and it would be easier to review if it is just whatever the siteimprove missed.

https://github.com/pod4lib/aggregator/blob/main/spec/features/accessibility_spec.rb

dnoneill:
closes #1261

### DIFF
--- a/app/components/dashboard/uploads_tab_component.html.erb
+++ b/app/components/dashboard/uploads_tab_component.html.erb
@@ -33,7 +33,7 @@
               <% status = files_status(uploads[0]) %>
               <%= render JobStatusIconComponent.new(status: status) %>
             <% else %>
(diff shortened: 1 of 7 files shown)


dnoneill:
@corylown I am not fully clear on the suggestion here. Just moving  <% unless show_label %>role="img" aria-label="<%= label %>"<% end %>></i> as a function or moving the entire tag into the rb file?

dnoneill:
@corylown Okay that is done. You were right it is much cleaner and allowed inheritance on the other icons. I also found a bug for the key on the organization upload page which got fixed.

corylown (approved):

PR #1323, opened 2025-10-23
Add filters to upload table for organizations

### LINKED ISSUE #1262: Create uploaded files filter
Right now you can just see a list. It would be nice if the icons/keys were clickable and the user could filter the results to just show for example the invalid marc files, etc.

Image

marlo-longley:
Closes #1262 

You can view this at URLs like `/organizations/cornell/streams/48d9316f-85fc-4e38-8dcb-b22f1407301c`

Because the `views/uploads/_file_table.html.erb` is used by two different controllers / show pages -- uploads and streams -- this got a little weird. Both views want to render the table, but the internal paths for filtering are different -- and in the case of `/uploads/:id` we don't even want the filtering.

### DIFF
--- a/app/assets/stylesheets/files.css
+++ b/app/assets/stylesheets/files.css
@@ -45,3 +45,7 @@
 .box-metadata {
   padding-right: 1rem;
 }
+
+#uploaded-file-filters > li.list-inline-item.active {
+   font-weight: bold;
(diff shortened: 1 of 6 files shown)


marlo-longley:
Closing in favor of #1429

PR #1326, opened 2025-10-23
Schedule stream compaction.

cbeer:


app/jobs/compact_uploads_job.rb
```diff
@@ -5,6 +5,22 @@
 class CompactUploadsJob < ApplicationJob
   queue_as :default
 
+  def self.enqueue_some(ratio: 0.25, maximum: 5) # rubocop:disable Metrics/AbcSize
+    streams = Stream.include(:statistic).default.select do |stream|
```
corylown: I think this should be `includes`?
```suggestion
    streams = Stream.includes(:statistic).default.select do |stream|
```

app/jobs/compact_uploads_job.rb
```diff
@@ -5,6 +5,22 @@
...
+      next unless stream.statistic.unique_record_count && stream.statistic.record_count
```
corylown: Should this be `stream.statistic&.unique_record_count` in case `stream.statistics` is `nil`?

corylown (approved):

PR #1366, opened 2025-10-29
Refactor upload status logic

marlo-longley:


corylown (commented):
It looks like all the various status pages are now only showing upload status, but the legend still makes it look like a whole bunch of statuses are possible that I think are not.
Image

There are different icon sizes displaying in the dashboard summary -- maybe because of the difference between there are no uploads and an upload was bad?
Image

PR #1373, opened 2025-10-31
Separate token and user ability rules into separate classes

corylown:
Maybe this will make adding _even more_ ability rules set by organizations somewhat more palatable.

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -1,80 +1,87 @@
 # frozen_string_literal: true
 
-# :nodoc:
+# Ability class that defines user-based abilities
 class Ability
   include CanCan::Ability
 
   attr_reader :allowlisted_jwt, :user
 
-  # rubocop:disable Metrics/CyclomaticComplexity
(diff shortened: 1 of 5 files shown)


app/models/ability.rb
```diff
@@ -1,80 +1,13 @@
...
-      can :read, MarcRecord, upload: { organization: { public: true } }
-      can %i[read profile normalized_data processing_status], Stream, organization: { public: true }
-      can %i[read info], Upload, organization: { public: true }
-      can :read, :pages_data
-      can %i[read users organization_details provider_details], Organization, public: true
+  def self.new(user, token = nil)
+    if token.present?
+      Abilities::TokenAbility.new(token)
+    else
+      Abilities::UserAbility.new(user)
```
cbeer: I haven't thought about this PR much, but overriding `new` feels a little funny. Maybe this condition is better in https://github.com/pod4lib/aggregator/blob/main/app/controllers/concerns/custom_pod_ability_concern.rb#L7?

PR #1414, opened 2025-11-05
Remove upload summary.

cbeer:
I can't remember if we agreed to remove the chronological list of uploads, or do something else to it..


app/components/dashboard/summary_component.html.erb
```diff
@@ -1,65 +1,60 @@
...
-      </div>
+    <!-- Normalized Data tab -->
+    <div id="normalized-data-pane" class="tab-pane fade" role="tabpanel">
+      <%= helpers.turbo_frame_tag "normalized_data_tab", src: dashboard_tab_path(:normalized_data), loading: "lazy" do %>
+        <div class="text-muted">Loading Normalized Data…</div>
+      <% end %>
     </div>
-  <% end %>
 
-  <%= render Dashboard::UploadsActivityComponent.new(uploads: @uploads) %>
```
corylown: I don't think this component is used anywhere else. Should we remove it too?

PR #1425, opened 2025-11-05
simplify dashboard routes

### LINKED ISSUE #1421: Can the route /dashboard/summary be simplified?
There seems to be no reason why it is a nested route. Can it be /dashboard or /summary?

dnoneill:
closes #1421


config/routes.rb
```diff
@@ -35,8 +35,8 @@ def matches?(request)
   get '/.well-known/resourcesync/capabilitylist', to: 'resourcesync#capabilitylist', as: :resourcesync_capabilitylist, defaults: { format: :xml }
   get '/.well-known/resourcesync/normalized-capabilitylist/:flavor', to: 'resourcesync#normalized_capabilitylist', as: :resourcesync_normalized_dump_capabilitylist, defaults: { format: :xml }
 
-  get 'dashboard/summary', to: 'dashboard#summary', as: :activity
-  get 'dashboard/tabs/:tab', to: 'dashboard#tab', as: :dashboard_tab
+  get 'dashboard', to: 'dashboard#summary', as: :activity
```
marlo-longley: would it make sense to have `#summary` actually be the index route? In the controller.

config/routes.rb
```diff
@@ -35,8 +35,8 @@ def matches?(request)
   get '/.well-known/resourcesync/capabilitylist', to: 'resourcesync#capabilitylist', as: :resourcesync_capabilitylist, defaults: { format: :xml }
   get '/.well-known/resourcesync/normalized-capabilitylist/:flavor', to: 'resourcesync#normalized_capabilitylist', as: :resourcesync_normalized_dump_capabilitylist, defaults: { format: :xml }
 
-  get 'dashboard/summary', to: 'dashboard#summary', as: :activity
-  get 'dashboard/tabs/:tab', to: 'dashboard#tab', as: :dashboard_tab
+  get 'dashboard', to: 'dashboard#index', as: :activity
```
corylown: I think an index action is needed in the controller for this to work, right?

dnoneill: Yeah that was my bad. I wasn't paying enough attention during rebase. Fixed.

PR #1426, opened 2025-11-05
Add Upload column to file table

marlo-longley:
This is in my notes and broken out from prior big PR -- but not sure if still desired
 
Screenshot 2025-11-05 at 1 17 16 PM


corylown (commented):
I think this was already addressed by https://github.com/pod4lib/aggregator/commit/14fb988b5435f997dedd7c813310f9bf4d358a58

PR #1430, opened 2025-11-05
Rename dashboard/summary to Activity

### LINKED ISSUE #1427: Consolidate "dashboard" "activity" and "summary"
I am hoping to consolidate the wording of "activity" "dashboard" and "summary" as they are all used for the same thing across the app. For example in the `en` file everything is in `dashboard.summary`, the link and route are aliased to "Activity" and the controller is named "dashboard".

marlo-longley:
Closes #1427

I find this less way confusing but tell me why it's a bad idea!


corylown (commented):
I like this, but should we also change the path/module name for the components from Dashboard to Activity? There's also a view path `/dashboard` that could also be renamed. What do you think?

marlo-longley:
@corylown thank you, I changed the folder names as you suggested though I didn't see any /dashboard view folder?

Also I am not sure in the diff why the `dashboard/users_tab_component.html.erb` doesn't appear like its siblings to simply have been renamed. Instead it shows as deleted/added. I followed the same method for moving all of those components.

PR #1434, opened 2025-11-06
Add a little more information to the marc records listing.

cbeer:
The checksum + index are maybe useful for identifying duplicate records across files 🤷‍♂️

### DIFF
--- a/app/views/marc_records/index.html.erb
+++ b/app/views/marc_records/index.html.erb
@@ -24,7 +24,10 @@
     <table class="table table-striped organizations mt-4">
       <thead>
         <tr>
-          <th>marc001</th>
+          <th style="width: 2rem">Type</th>
+          <th>MARC 001</th>
+          <th>Index</th>
+          <th>Checksum</th>
           <th>Download</th>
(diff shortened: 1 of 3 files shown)


app/views/marc_records/index.html.erb
```diff
@@ -24,22 +24,30 @@
...
       </thead>
 
       <tbody>
         <% @marc_records.each do |marc_record| %>
           <tr data-turbo="false">
+            <td><%= marc_record.delete? ? tag.i(class: "bi bi-minus text-danger", title: 'Delete') : tag.i(class: "bi bi-plus text-success", title: 'Add/update') %></td>
             <td><%= marc_record.marc001 %></td>
-            <td><%= link_to marc_record.file.filename, file_info_organization_upload_path(@organization, marc_record.upload, marc_record.file) %></td>
+            <td><%= marc_record.index %></td>
+            <td class="small"><%= marc_record.checksum[-8..] %></td>
```
corylown: This won't work with records from deletes where I think checksum is nil.

corylown (approved):

Issue #1291, opened 2025-10-20
Allow POD contributors to control access to their records


corylown:
I'm making a number of assertions based on various conversations and emails about the behavior and requirements of this functionality. These are intended to convey my best guess at the functionality required, but at least convey the level of detail needed to implement these features.

1. In order to access records contained in POD a user must have a POD account.
2. Users designated as POD site admins have full access to all records in POD.
3. Unless set otherwise, all POD orgs and groups have access to an organization's records.
4. Organization owners can control which organizations or groups can access their records.
5. POD site admins can control which organizations or groups can access any organization's records.
6. An organization can set access restrictions at the group or organization level.
7. If group access is granted all organizations in that group and their members have access to the organization's records regardless of whether organization level access is granted. (Group access takes precedence over organization access.)
8. All logged in users can see all organizations in POD regardless of whether they are permitted to access that organization's records.
9. The POD UI will provide POD admins with a comprehensive view showing which organizations and groups have access to each organization's records in POD.
10. The POD UI will provide organization owners and members with a view that shows which organizations and groups have access to their own organization's records in POD.

corylown:
We're going to approach this as a spike for testing on stage. Until we learn more we will either not merge this feature into main or keep it hidden behind a feature flag.

JohnMarkOckerbloom:
1. I think it's sufficient to deny access to the record data in this case.  I don't think they have to be denied access to references such as IDs (though they don't have to be given access to them either.)
2. I think what you're describing is fine.  Access control ultimately reflects the policy of institutions, rather than the policy of groups.

corylown:
Feedback and todo:

- Change "Are you sure?" confirmation dialog text to reflect the actual change about to be made. "Are you sure you want to grant Stanford access?"
- Consult with Astrid about how to better layout/draw attention to the access management editing part of the UI that is currently at the very bottom of the page.
- To the reason column (why an institution has access) add reference to both the organization or group or the most specific kind of access granted (organization).
- When attempting to download a file you do not have access to the proxy controller currently returns a server error instead of a 403.

PR #1483, opened 2025-11-13
Add restrict_downloads field to organizations to limit access to uploads/records

corylown:
Part of #1291 

Another step toward organization controlled download restrictions. Tom wants the default to be restricted so that's what I've done, but to maintain the current behavior for existing organizations I flip this to unrestricted in the initial migration. I added a control in the UI for POD site admins to change this restriction setting. We will likely want to open this up to org owners once more fine-grained access controls are added.

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -34,11 +34,15 @@ class Ability
   def user_with_roles_abilities
     return if user.roles.empty?
 
-    can :read, ActiveStorage::Attachment
-    can :read, MarcRecord
-    can :read, Stream
-    can :read, Upload
     can :read, :pages_data
+    organization_ability_restrictions
+  end
+
+  def organization_ability_restrictions(restrictions = { restrict_downloads: false })
+    can :read, ActiveStorage::Attachment, { record: { organization: restrictions } }
+    can :read, MarcRecord, upload: { organization: restrictions }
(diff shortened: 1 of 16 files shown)


app/models/ability.rb
```diff
@@ -34,11 +34,15 @@ def all_user_abilities
...
-    can :read, Upload
     can :read, :pages_data
+    organization_ability_restrictions
+  end
+
+  def organization_ability_restrictions(restrictions = { restrict_downloads: false })
+    can :read, ActiveStorage::Attachment, { record: { organization: restrictions } }
+    can :read, MarcRecord, upload: { organization: restrictions }
+    can :read, Stream, organization: restrictions
+    can :read, Upload, organization: restrictions
```
cbeer: It took me a moment to figure out what was going on with the `restrictions` argument here. I guess I'm hoping that being this clever pays off in the next step (when we're also allowing members of organizations that have allowed access etc..), but maybe it'd be clearer if we duplicated the permissions (in different methods, e.g. `def allow_read_access_to_unrestricted_organization_data`, `def allow_read_access_to_organization_data_by_organization_ids(ids)`).

But I'm happy to just be assured it'll all make total sense in a bit 🤷

corylown: I may get to the next step and find I agree with you 😸

Issue #1482, opened 2025-11-13
Make marc profile documentation link clickable for admins and org owners


corylown:
It's a little annoying that if you have the ability to edit the provider details there's no way to click the marc documentation link on the provider details pages: `/organizations/jhu/provider_details`. Maybe we could display the current value as a link above the form?

<img width="969" height="835" alt="Image" src="https://github.com/user-attachments/assets/8fc7a295-7510-4917-b811-d43ba0bd2b42" />

PR #1490, opened 2025-11-13
Link out to MARC documentation for admins

marlo-longley:
Closes [#1482](https://github.com/pod4lib/aggregator/issues/1482)

@corylown since the link is already active on non-admin views (so a link above the form would be redundant?) what about something like this?

Admin view
Screenshot 2025-11-13 at 12 03 12 PM

Non admin view
Screenshot 2025-11-13 at 12 03 20 PM

### DIFF
--- a/config/brakeman.ignore
+++ b/config/brakeman.ignore
@@ -1,0 +1,49 @@
+{
+  "ignored_warnings": [
+    {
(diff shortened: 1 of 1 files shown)


corylown (commented):
Works for me.

corylown (approved):

PR #1498, opened 2025-11-14
Create a Downloader model to manage access to organization records/files

corylown:
Part of #1291

### DIFF
--- a/app/models/downloader.rb
+++ b/app/models/downloader.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+# Class representing a downloader relationship between an organization and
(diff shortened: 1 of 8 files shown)


app/models/downloader.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+# Class representing a downloader relationship between an organization and
+# resources (a group or another organization) that are allowed to download from it
+class Downloader < ApplicationRecord
+  has_paper_trail
```
corylown: I'm not sure I need this because we have papertrail on orgs and groups.

PR #1506, opened 2025-11-19
Allow non site admins to view current access settings

corylown:
Part of #1291 

Adds a view of the current access settings for non-admins:

Screenshot 2025-11-19 at 1 19 14 PM

### DIFF
--- a/app/components/downloaders/administer_access_component.html.erb
+++ b/app/components/downloaders/administer_access_component.html.erb
@@ -1,16 +1,11 @@
 <h4 class="mt-4"><%= t('downloaders.administer_access_component.heading') %></h4>
-<% unless @organization.restrict_downloads %>
-  <div class="alert alert-warning" role="alert">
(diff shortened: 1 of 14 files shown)


app/components/downloaders/administer_access_component.html.erb
```diff
@@ -1,16 +1,11 @@
 <h4 class="mt-4"><%= t('downloaders.administer_access_component.heading') %></h4>
-<% unless @organization.restrict_downloads %>
-  <div class="alert alert-warning" role="alert">
-    <%= t('downloaders.administer_access_component.unrestricted_alert') %>
-  </div>
-<% end %>
```
cbeer: I'm assuming you've intentionally dropping this banner because the icons accurately reflect the actual state?

corylown: Yup. And I'm about to put up another PR with a more helpful message.

PR #1509, opened 2025-11-19
Add a summary table to detail download access

corylown:
Part of #1291 

Adds a tabbed interface to see a summary of access both granted and received for an organization.
Screenshot 2025-11-19 at 4 44 40 PM

It would be nice if this updated live as changes were made, but it's a bit fiddly and I'd rather address that as follow up work.

### DIFF
--- a/app/components/downloaders/accessible_summary_table_data_component.rb
+++ b/app/components/downloaders/accessible_summary_table_data_component.rb
@@ -1,0 +1,46 @@
+# frozen_string_literal: true
+
+module Downloaders
+  # Table data component that shows which organizations this organization can access
+  class AccessibleSummaryTableDataComponent < AccessSummaryTableDataComponent
+    def explanation_text
+      unless @other_org.provider?
+        return I18n.t('downloaders.accessible_summary_table_data_component.explanation.not_a_provider',
+                      grantor_name: grantor_name)
+      end
+
+      super
+    end
+
+    private
+
+    def grantee_name
+      @organization.name
+    end
+
+    def grantor_name
+      @other_org.name
+    end
+
+    def group_memberships_text
+      (@other_org.downloader_groups & @organization.groups).map(&:display_name).to_sentence
+    end
+
+    def can_download?
+      @other_org.provider? &&
+        (unrestricted_access? || @organization.effective_downloadable_organizations.include?(@other_org))
+    end
+
+    def unrestricted_access?
+      !@other_org.restrict_downloads
+    end
+
+    def access_granted_via_group_membership?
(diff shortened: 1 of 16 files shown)


app/views/downloaders/index.html.erb
```diff
@@ -3,5 +3,30 @@
...
+    <%# Manage access tab %>
+    <div class="tab-pane fade show active" id="manage-pane" role="tabpanel" aria-labelledby="manage-tab">
+      <%= render Downloaders::AdministerAccessComponent.new(organization: @organization) %>
+    </div>
+
+    <%# Summary of access tab %>
+    <div class="tab-pane fade" id="access-pane" role="tabpanel" aria-labelledby="access-tab">
+      <%= render Downloaders::AccessSummaryTableComponent.new(organization: @organization) %>
+    </div>
+  </div>
```
cbeer:
```suggestion
  <turbo-frame id="access-frame">
    <div class="tab-content" id="downloaders-tabs-content">
      <%# Manage access tab %>
      <div class="tab-pane fade show active" id="manage-pane" role="tabpanel" aria-labelledby="manage-tab">
        <%= render Downloaders::AdministerAccessComponent.new(organization: @organization) %>
      </div>

      <%# Summary of access tab %>
      <div class="tab-pane fade" id="access-pane" role="tabpanel" aria-labelledby="access-tab">
        <%= render Downloaders::AccessSummaryTableComponent.new(organization: @organization) %>
      </div>
    </div>
  </turbo-frame>
```

Could we (temporarily?) move the `turbo-frame` from `AccessCheckboxComponent` to wrap both pieces of tab content? I suspect this isn't as-ideal (and might require reverting if/when we use some turbo streams), but at least it's less confusing for the moment.

app/components/downloaders/accessible_summary_table_data_component.rb
```diff
@@ -0,0 +1,46 @@
...
+    end
+
+    def can_download?
+      @other_org.provider? &&
+        (@other_org.restrict_downloads == false || @organization.effective_downloadable_organizations.include?(@other_org))
+    end
+
+    def unrestricted_access?
+      @other_org.restrict_downloads == false
+    end
```
cbeer: I'm not loving `== false`, but maybe there's a reason to distinguish between nil + false?


```suggestion
    def can_download?
      @other_org.provider? &&
        (unrestricted_access? ||
          @organization.effective_downloadable_organizations.include?(@other_org))
    end

    def unrestricted_access?
      !@other_org.restrict_downloads
    end
```

PR #1511, opened 2025-11-20
Consolidate AccessibleSummaryTableDataComponent into AccessSummaryTableDataComponent

cbeer:
I think the output is the basically the same, besides swapping the org/other org around, maybe?


app/components/downloaders/access_summary_table_data_component.rb
```diff
@@ -34,31 +37,31 @@ def icon
...
     end
 
     def group_memberships_text
-      (@organization.downloader_groups & @other_org.groups).map(&:display_name).to_sentence
+      (@grantor_org.downloader_groups & @grantee_org.groups).map(&:display_name).to_sentence
     end
 
     def can_download?
-      unrestricted_access? || @other_org.effective_downloadable_organizations.include?(@organization)
+      unrestricted_access? || @grantee_org.effective_downloadable_organizations.include?(@grantor_org)
```
corylown: Since we always show "ORG is not a provider" for non-providers, should we also check for that here in the weird case where somehow a non-provider has restrict_download flipped to false? Probably wouldn't/shouldn't happen...

cbeer: Yeah, that's a weird case, I think. Should we suppress the whole column (and access controls) for non-providers (64f6bca)?

corylown (approved):

